PR #11, opened 2021-07-09
Added support for synthesizing system verilog designs with vivado

jaromharris:
I did this because I wanted to run my ECEN 220 and ECEN 323 labs through bfasst. Do you think this is something we should add to bfasst?


jgoeders:
Sure, SystemVerilog support is great.  And adding your 220/323 designs.

PR #23, opened 2022-04-28
Added BYU folder into examples with 8to1 Mux, ALU and UART skeletons

TheDHCreator:
Skeletons (mostly empty .v files and complete yaml files) have been made for all next designs. The UART you put into slack is there (needs to be converted into verilog from system verilog) and everything else is there. Also added yaml files for the and2 designs in basic (I thought you wanted me to delete those but for some reason they're back now).


jgoeders:
Can you delete the and2 designs?  We don't need and2 and and3 designs.

jgoeders:
Does the uart pass the `xilinx` flow?  I don't think we need to create PRs to add benchmarks until they at minimum pass the `xilinx` flow.

PR #27, opened 2022-04-28
AND2s and FF gone once and for all, experiments modified with the new BYU folder

TheDHCreator:
Turns out the reasons the AND2 designs and the FF keep reappearing is because they are currently on the master branch still, that's why Pablo and Jake still have them. So, I've made a new branch where I'll work on my designs, I've removed those other designs, and modified the verify experiment yaml files to include our new BYU designs folder. Please merge as soon as you can. Thanks!


experiments/verify_fasm_to_bels.yaml
```diff
@@ -1,6 +1,15 @@
...
-    # - base/
+    - byu/
+    #- ipassurance_designs/
+    #- lfsr_randgen/
+    #- memories/
+    #- misc/
+    #- ooc/
+    #- riscv/
+    #- simple/
+    #- uart/
```
jgoeders, line 12: Can you remove these commented out lines?

experiments/verify_fasm_to_bels_yosys.yaml
```diff
@@ -1,5 +1,14 @@
...
-    # - base/
+    - byu/
+    #- ipassurance_designs/
+    #- lfsr_randgen/
+    #- memories/
+    #- misc/
+    #- ooc/
+    #- riscv/
+    #- simple/
+    #- uart/
```
jgoeders, line 12: Remove

jgoeders (changes requested):
Can you leave the FF files in? I don't want to remove those. 

The idea of the basic designs is to exercise the basic FPGA primivites. The and2 and and3 test the same thing (single LUT) so we don't need both. But having the FF design is good. 

You don't need to make a new PR to address this. Just push again to this branch.

experiments/verify_fasm_to_bels.yaml
TheDHCreator: Yes I'll do that.

PR #29, opened 2022-04-29
FF design files restored, AND2 removed once and for all, SS Display Design now available and working!

TheDHCreator:
This is a merge of a few commits from yesterday (accidentally removed FF files and they have since been put back) and I removed the AND2 designs finally. I also took my Seven Segment Display from 220, converted it to Verilog, and it ran through and passed equivalence checking on both flows! Woohoo!


experiments/verify_fasm_to_bels.yaml
```diff
@@ -1,7 +1,7 @@
 designs:
+    - base/
```
jgoeders, line 2: I don't know if everyone in the base folder compares as equivalent.  Does it?  Let's only include things that come back as equivalent.

experiments/verify_fasm_to_bels_yosys.yaml
```diff
@@ -1,5 +1,8 @@
 designs:
+    - base/
```
jgoeders, line 2: I don't know if everyone in the base folder compares as equivalent.  Does it?  Let's only include things that come back as equivalent.

jgoeders (changes requested):

experiments/verify_fasm_to_bels_yosys.yaml
TheDHCreator: Okay! I will only include the ones that are currently equivalent

TheDHCreator:
Just committed the changes that should satisfy your requests. Take a look when you can and let me know if you want any more changes. Thanks for a great week! See you Monday!

TheDHCreator:
Just added the change to vivado.py in scripts/bfasst/synth before you merge it.

examples/basic/rom/rom.v
```diff
@@ -1,59 +1,55 @@
-// ROMs Using Block RAM Resources.
```
jgoeders, line 1: What's the reasoning for changing this file?  Does this new file use BRAMs?

TheDHCreator: I got a newer ROM design working, so I decided to push it as well. Do I check the netlists to see if it uses BRAMs?

jgoeders: Yes, check the nestlist.

TheDHCreator: It currently doesn't look like it's using BRAMs. Just I and OBUFs. Should I get rid of the design and commit before you merge?

TheDHCreator:
ROM design is back to the old one that uses BRAMs now.

jgoeders (approved):

PR #32, opened 2022-05-03
Jake designs

JakeEdvenson:
Fixed the ReadMe based upon the feedback you gave me, fixed the conformal.py to automatically create folders that typically give errors, and created a oneshot.v file that passes off both flows.


README.md
```diff
@@ -18,3 +18,28 @@ usage: run_experiment.py [-h] [-j THREADS] experiment_yaml
...
+
+* Use ```make rapidwright```
+
+* Use ```make env```
+
+* Use ```source env.sh```
+
+* Use ```make install``` 
+
+Finally, test to confirm that everything worked correctly! Run ```source env.sh``` and ```./scripts/run_experiment.py experiments/verify_fasm_to_bels.yaml``` from the bfasst directory to check if everything works correctly.
```
jgoeders, line 37: Let's just use `run_design` here instead of `run_experiment`.

README.md
```diff
@@ -18,3 +18,28 @@ usage: run_experiment.py [-h] [-j THREADS] experiment_yaml
 run_experiment.py: error: the following arguments are required: experiment_yaml
 ```
 
+**How to Install BFASST** 
+
+* Clone the github repository into your home directory in linux. Repository is found at https://github.com/byuccl/bfasst. Use ```git clone https://github.com/byuccl/bfasst.git.``` (Note: if Git is not installed on your device, use ```sudo apt install git``` first.)
```
jgoeders, line 23: Let's remove the part about the home directory.  The code should work wherever you clone it.

scripts/bfasst/compare/conformal.py
```diff
@@ -74,6 +74,19 @@ def compare_netlists(self, design, print_to_stdout=True):
         # Create do file
         do_file_path = self.create_do_file(design)
 
+        # Create remote machine folders
+        cmd = (
+            "mkdir -p bfasst_libs;"
+            + "mkdir -p bfasst_libs/xilinx;"
+            + "touch bfasst_libs/xilinx/cells_sim.v;"
```
jgoeders, line 81: I don't think you need this touch command.  Can you verify it works without it?

scripts/bfasst/compare/conformal.py
```diff
@@ -74,6 +74,19 @@ def compare_netlists(self, design, print_to_stdout=True):
...
+            + "mkdir -p bfasst_work;"
+            + "touch bfasst_work/compare.do;"
```
jgoeders, line 83: I don't think you need this touch command. Can you verify it works without it?

README.md
jgoeders: Also should this be run using the python virtual environment instead of `./`?

jgoeders:
OK, I've added some more comments.  In the future, just use your existing pull request as it tracks my comments and what I have already looked at.  There's no need to close the pull request and make a new one. 

Thanks!

README.md
JakeEdvenson: Ok, I have it defaulting to run in a python virtual environment and running the check on an add4.

scripts/bfasst/compare/conformal.py
JakeEdvenson: Both work without the touch command.

JakeEdvenson:
Okay, I just pushed a new commit that should've fixed everything!

jgoeders (approved):

PR #34, opened 2022-05-05
Some new example files w/ size descriptions

JakeEdvenson:
Added buttoncount, calc, regfile, shiftReg, and UpDownButtonCount. Currently, 3/5 work with Yosys. All work with Conformal. It is marked at the top of both buttoncount and regfile that they are not working with Yosys yet.


jgoeders:
Looks good, but rather than adding statuses and resource counts in the actual source files, could you instead create a markdown file with a table listing this information?  Thanks!

jgoeders (approved):

PR #48, opened 2022-05-26
Updated Logic for bit_num_reg_2 netlist

palm9727:
The circuit for the transmitter's second non-equivalent FF on its own is equivalent.


palm9727:
The circuit for the transmitter's third non-equivalent FF on its own is also equivalent.

PR #51, opened 2022-08-17
Added netlist mapping and flow arguments to bfasst

palm9727:


scripts/run_design.py
```diff
@@ -51,9 +52,12 @@ def main():
     # Get the flow object
     flow = flows.Flows(args.flow)
 
+    # Get the flow arguments
+    flow_args = args.flow_args
```
jgoeders, line 56: I think we talked about making flow_args a dictionary so that the code here would look something like this:
```
flow_args[SYNTH_STAGE] = args.synth_args
flow_args[IMPL_STAGE] = args.impl_args
...
```

scripts/bfasst/netlist_mapping.py
```diff
@@ -0,0 +1,321 @@
+# Structural Mapping Algorithm developed for the bfasst project in the BYU CCL
```
jgoeders, line 1: Can you format this with `black`?

scripts/bfasst/netlist_mapping.py
```diff
@@ -0,0 +1,321 @@
...
+        self.other_wires_names = other_wires_names
+        self.other_wires_number = other_wires_number
+
+def generate_conformal_mapped_points_file(top_instance, golden_module_name, reversed_module_name):
+    # Loop through each of the ports in the instance
+    for top_port in top_instance.get_ports():
+        if "IN" in str(top_port.direction):
+            if len(top_port.pins) > 1:
+                for pin in top_port.pins:
+                    if pin.wire != None:
```
jgoeders, line 35: Should be `is not None`

scripts/bfasst/netlist_mapping.py
```diff
@@ -0,0 +1,321 @@
...
+                        input_name = pin.wire.cable.name + "[" + str(pin.wire.index()) + "]"
+                        print("add mapped points " + input_name + " " + input_name 
+                              + " -type PI PI -module " + golden_module_name + " " + reversed_module_name)
+            else:
+                print("add mapped points " + top_port.name + " " + top_port.name + 
+                " -type PI PI -module " + golden_module_name + " " + reversed_module_name)
+        elif "OUT" in str(top_port.direction):
+            if len(top_port.pins) > 1:
+                for pin in top_port.pins:
+                    if pin.wire != None:
```
jgoeders, line 45: Same as above

scripts/bfasst/netlist_mapping.py
```diff
@@ -0,0 +1,321 @@
...
+                        print("add mapped points " + input_name + " " + input_name 
+                              + " -type PO PO -module " + golden_module_name + " " + reversed_module_name)
+            else:
+                print("add mapped points " + top_port.name + " " + top_port.name + 
+                " -type PO PO -module " + golden_module_name + " " + reversed_module_name)
+        else:
+            print("Unable to recognize port of the top module!")
+
+
+def generate_mapped_points_file(mapped_points, golden_module_name, reversed_module_name):
```
jgoeders, line 56: Can you add a docstring to your functions?

scripts/run_design.py
jgoeders: And it would be good to make it like this so that we don't get KeyErrors if a user doesn't provide arguments for a certain type:
```
from collections import defaultdict
flow_args = defaultdict(str)
```

scripts/bfasst/flows.py
```diff
@@ -61,14 +75,79 @@ def get_flow_fcn_by_name(flow_name):
...
+def run_flow(design, flow_type, flow_args, build_dir, print_to_stdout=True):
     assert type(design) is bfasst.design.Design
 
     flow_fcn = flows.get_flow_fcn_by_name(flow_type)
-    return flow_fcn(design, build_dir, print_to_stdout)
-
+    return flow_fcn(design, flow_args, build_dir, print_to_stdout)
+
+
+def parse_flow_args(flow_args):
```
jgoeders, line 85: Sorry, I'm not quite sure what this function is doing.  It looks like it's manually parsing the arguments? I'm not quite sure what it's for...

scripts/bfasst/flows.py
```diff
@@ -61,14 +75,79 @@ def get_flow_fcn_by_name(flow_name):
...
+            args[Args.IMPL_STAGE.value] = tools[i]
+        elif stage == "mapping":
+            args[Args.MAP_STAGE.value] = tools[i]
+        elif stage == "equivalence":
+            args[Args.EQUIV_STAGE.value] = tools[i]
+    
+    return args
 
-def flow_ic2_lse_conformal(design, build_dir):
+def flow_ic2_lse_conformal(design, flow_args, build_dir):
```
jgoeders, line 150: This is correct, to pass `flow_args` into each flow.  But then within the flow, we want to pass along the option strings into each tool.  

For example:
```
synth_tool = bfasst.synth.ic2_lse.IC2_LSE_SynthesisTool(build_dir, flow_args[SYNTH_STAGE])
```

Then within `tool.py`, at the constructor: https://github.com/byuccl/bfasst/blob/4081d15f33a9a95c4c33e18b84eb8a466a3b0607/scripts/bfasst/tool.py#L19 you would want to accept the argument string for this tool, and use argparse to parse it into a `self.args` that you can then reference for the individual tool.

Does that make sense?

scripts/bfasst/compare/conformal.py
```diff
@@ -25,14 +26,16 @@ class Conformal_CompareTool(CompareTool):
...
+    MAPPED_POINTS_FILE_NAME = "mapped_points.txt"
 
     def __init__(self, cwd, vendor):
         super().__init__(cwd)
 
         assert type(vendor) is flows.Vendor
         self.vendor = vendor
 
-    def compare_netlists(self, design, print_to_stdout=True):
+    def compare_netlists(self, design, mapping_algorithm, print_to_stdout=True):
```
jgoeders, line 38: See my comments below.  I don't think we want to change the argument list here. Rather, if using the approach described below, you should have access to `self.args` and can use that to get a mapping algorithm option.

scripts/bfasst/flows.py
jgoeders: Although you can add the `args` argument to the base `Tool` class, the actual logic for the argument parsing will be depending on the type of tool, so the argparse code would actually be placed in the base class for each tool type, such as in the `__init__` of `SynthesisTool`.

reillymck:
The dictionary implementation makes sense. Side question: why in scripts/bfasst/compare/conformal.py line 93 are we using os.system to call a python process? Why aren't we just importing the appropriate python function? Or at least using Popen()? Also, line 83 incorrectly assumes the user's location of the bfasst project. Is there already a global variable in the config/environment variable to set? Although if we just import the python function this specific instance is irrelevant.

jgoeders:
Those both sound like they slipped past my code reviews.  Yes, please fix it to use subprocess instead of os, and please using the paths in paths.py instead of hard-coded paths.

scripts/bfasst/compare/conformal.py
```diff
@@ -72,7 +75,24 @@ def compare_netlists(self, design, print_to_stdout=True):
...
+                + self.MAPPING_FILE_NAME 
+                + " "
+                + str(design.impl_netlist_path)
+                + " "
+                + str(design.reversed_netlist_path)
+                + " > "
+                + self.MAPPED_POINTS_FILE_NAME)
+        else:
+            command = "echo No BYU CCL Mapping > mapped_points.txt"
+        os.system(command)
```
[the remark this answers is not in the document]
reillymck: Done. Currently the python function is called directly; no need for a bash call.

scripts/bfasst/netlist_mapping.py
reillymck: done

scripts/bfasst/netlist_mapping.py
reillymck: done

PR #54, opened 2022-08-30
Added parser. Reorganized files.

JakeEdvenson:
Updated the file structure so now the waveform viewer and the diff.txt parser are not in the compare directory (so only the actual comparison tools like waveform.py, yosys.py, and conformal.py are in the folder.) Also updated the parser so it should work with most files. I've tested it on several and it seems to be working well. It's now been integrated into the comparison process.


scripts/bfasst/compare_waveforms/analyze_graph.py
```diff
@@ -21,14 +21,10 @@ def analyze_graphs(path, module):
     impl_fst = path / Path(module + "_impl.vcd.fst")
```
jgoeders, line 21: If `path` is already a Path object, you don't need to create another Path object.  For example, this code works and is cleaner:
```
impl_fst = path / module + "_impl.vcd.fst"
```

jgoeders (approved):

scripts/bfasst/compare_waveforms/analyze_graph.py
jgoeders: @JakeEdvenson

PR #60, opened 2022-09-01
Fixed Vivado Automation Issues

JakeEdvenson:
Because the modules are named differently in the impl.v and reversed.v files, I added in logic to rename them to names that the testbench is looking for. Now when Vivado launches, it can link the two files together and automatically start the simulation tool with the testbench's simulation data.


jgoeders:
@JakeEdvenson This has a conflict.

jgoeders (approved):

PR #62, opened 2022-09-07
Automatic gtkwave closing & Code Refactoring

JakeEdvenson:
During equivalence checking, gtkwave now closes itself automatically so the user doesn't have to close it to find out if the graphs are equivalent. If the user wants to view the graphs, they just have to re-run the program and they can use the tools from analyze_graph to do so. 

Also, when making my graph I noticed that several files are generated in the equivalence checking portion of the script, so I created their own functions and call them during the file generation portion of the script. 

Finally, on the code cleanup side, I'm turning all of the paths into one struct and passing this struct into each function rather than having paths being declared all over the place. I'm also re-factoring the code so that most functions are in dedicated files in the compare_waveforms directory. As I go through, I'm going to be adding commentary to make everything more clear for me and for anyone using this tool.


scripts/bfasst/compare/waveform.py
```diff
@@ -1,51 +1,31 @@
...
+    file_rewriter,
+)
+from bfasst.compare_waveforms.Templates import get_paths
+from bfasst.compare_waveforms.Interface import waveform_interface
+
+# Data and Paths are structs that contain the parsed data from our design and the paths for all generated files.
+data = parse_files.data
+paths = get_paths.paths
 
 """The main class for comparing the waveforms."""
```
dallinjdahl, line 30: Docstrings should be the first statement within a scope, not just before.

scripts/bfasst/compare/waveform.py
```diff
@@ -73,536 +53,121 @@ def compare_netlists(self, design):
...
-            for number, line in enumerate(lines):
-                if number not in [2,3]:
-                    fp.write(line)
-        return is_equivalent
+    def run_test(self):
+        return parse_diff.check_diff(
+            paths
+        )  # Checks the two VCD files against each other. Returns either equivalent or not depending on how many lines are different.
 
     def check_compare_status(self, log_path):
```
dallinjdahl, line 172: This function should have a default return value.

scripts/bfasst/compare_waveforms/File_Generation/tcl_generator.py
```diff
@@ -0,0 +1,38 @@
+"""Generates the first TCL that will be used in gtkwave to create a VCD output."""
```
dallinjdahl, line 1: Docstring should be within function.

scripts/bfasst/compare/waveform.py
```diff
@@ -73,536 +53,121 @@ def compare_netlists(self, design):
...
-            cont = input("Warning: Detected multiple verilog files. Results may be inaccurate.\n Continue anyways? 1 for yes, 0 for no.")
-            if(cont == "0"):
-                print("Ok. Returning non-equivalent status.")
-                return Status(CompareStatus.NOT_EQUIVALENT)
-
-        #This series of if/else statements is used to check if the tests have already been performed. If they have, an option is presented
-        # for the user to view the previously-generated waveforms. If the design was unequivalent previously, the diff output will be
-        # pasted into the linux terminal so they can view specific times in the waveform.
+        paths = get_paths.get_paths(
+            self, design
```
dallinjdahl, line 57: self should never be passed explicitly.  If a function requires self, it should be part of the class.  In addition, it looks like paths is dependent on the filepaths defined in self, so that should also be made part of the class.

scripts/bfasst/compare_waveforms/File_Generation/tcl_generator.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def generate_first_TCL(paths, data, i):
+    path = paths["build_dir"] / (paths["modules"][i + 1] + ".tcl")
+    if path.exists():
+        path.unlink()
+    with path.open("x") as TCL:
+        line = "set filter [list "
+        for totalData in data["total_list"]:
+            line = line + paths["modules"][i + 1] + "_tb." + totalData.strip() + " "
```
dallinjdahl, line 11: String generation should be done with f-strings, ie:
```python
line = f"{line}{paths[modules][i+1]}_tb.{totalData.strip()} "
```

scripts/bfasst/compare_waveforms/File_Generation/tcl_generator.py
```diff
@@ -0,0 +1,38 @@
...
+        line = line + "]\n"
+        TCL.write(line)
+        TCL.write("gtkwave::addSignalsFromList $filter\n")
+        TCL.write(
```
dallinjdahl, line 15: f-strings

scripts/bfasst/compare_waveforms/File_Generation/waveform_generator.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+    subprocess.run(["vvp", str(paths["dsn"])])
+    subprocess.run(["mv", "test.vcd", paths["temp_vcd"][i]])
+    paths["dsn"].unlink()
+
+    gtkwave(paths, i)
+
+
+"""Creates a waveform that only takes into account the inputs/outputs (totally ignores all internal components) and then
+takes the input/output graph and turns it into a VCD file to be compared against."""
```
dallinjdahl, line 28: Have you looked into using the $dumpvars(1, MODULENAME) syntax to limit iverilog's output to top level signals?
http://referencedesigner.com/tutorials/verilog/verilog_62.php

scripts/bfasst/compare_waveforms/File_Generation/testbench_generator.py
```diff
@@ -0,0 +1,141 @@
...
+                if "TB_NAME;" in line:
+                    line = "module " + paths["modules"][1] + "_tb;"
+                    tb.write(line)
+                    line = "\n"
+
+                if "TB_NAME)" in line:
+                    line = "    $dumpvars(0," + paths["modules"][1] + "_tb);\n"
+
+                if "INPUTS" in line:
+                    for input, bits in zip(data["input_list"], data["input_bits_list"]):
```
dallinjdahl, line 37: input also doubles as a builtin, probably don't want to shadow it.

scripts/bfasst/compare_waveforms/File_Generation/file_rewriter.py
```diff
@@ -0,0 +1,46 @@
...
+"""A function to fix syntax issues that spydrnet commonly has with xilinx-generated or reverse-generated netlists."""
+
+
+def fix_file(paths, i):
+    path = paths["file"][i]
+    file_name = paths["modules"][i + 1]
+    with path.open("r") as fin:
+        file_data = fin.read()
+        if i == 1:
+            file_data = file_data.replace("module top(", "module " + file_name + "(")
```
dallinjdahl, line 20: f-strings

scripts/bfasst/compare_waveforms/File_Generation/testbench_generator.py
```diff
@@ -0,0 +1,141 @@
...
+and replaces certain variables with the corresponding information from the data structure. It also sets the variables equal
+to random numbers that are generated to be within the corresponding variable's bit range."""
+
+
+def generate_first_testbench(paths, test_num, data, i):
+    with paths["sample_tb"].open() as sample:
+        with paths["tb"][i].open("x") as tb:
+            for line in sample:
+                if "TB_NAME;" in line:
+                    line = "module " + paths["modules"][1] + "_tb;"
```
dallinjdahl, line 29: f-strings

scripts/bfasst/compare_waveforms/File_Parsing/parse_files.py
```diff
@@ -0,0 +1,108 @@
...
+    for port in definition.ports:
+        if str(port.direction) == "Direction.OUT":
+            data["output_list"].append(port.name)
+            data["total_list"].append(port.name)
+            data["output_bits_list"].append(len(port.pins) - 1)
+        elif str(port.direction) == "Direction.IN":
+            data["input_list"].append(port.name)
+            data["total_list"].append(port.name)
+            data["input_bits_list"].append(len(port.pins) - 1)
+    if data["input_list"].__contains__("clk"):
```
dallinjdahl, line 27: I think this should be `in` rather than `__contains__`

scripts/bfasst/compare_waveforms/File_Parsing/parse_files.py
```diff
@@ -0,0 +1,108 @@
...
+
+    contains_item = False
+    not_port = False
+
+    for i in library.definitions:
+        for port in i.ports:
+            for item in total_reversed:
+                if item == port.name:
+                    contains_item = True
+            if contains_item == False:
```
dallinjdahl, line 59: should be `is False`.

scripts/bfasst/compare_waveforms/File_Parsing/parse_files.py
```diff
@@ -0,0 +1,108 @@
...
+                not_port = True
+            else:
+                contains_item = False
+        if not_port == False:
```
dallinjdahl, line 63: `is False`

scripts/bfasst/compare_waveforms/File_Parsing/parse_files.py
```diff
@@ -0,0 +1,108 @@
...
+                if str(port.direction) == "Direction.OUT":
+                    data["output_list"].append(port.name)
+                    data["total_list"].append(port.name)
+                    data["output_bits_list"].append(len(port.pins) - 1)
+                elif str(port.direction) == "Direction.IN":
+                    data["input_list"].append(port.name)
+                    data["total_list"].append(port.name)
+                    data["input_bits_list"].append(len(port.pins) - 1)
+
+            if data["input_list"].__contains__("clk"):
```
dallinjdahl, line 74: should use `in` instead of `__contains__`

scripts/bfasst/compare_waveforms/File_Parsing/parse_files.py
```diff
@@ -0,0 +1,108 @@
...
+                    newFile.write(line)
+                elif "output" in line:
+                    newFile.write(line)
+                else:
+                    if j == 0:
+                        j = 1
+                        newFile.write(line)
+    if i == 0:
+        return parse_multiple(paths["path"][i], paths["test"])
+    else:
```
dallinjdahl, line 107: unnecessary else

scripts/bfasst/compare_waveforms/File_Parsing/parse_files.py
```diff
@@ -0,0 +1,108 @@
+import spydrnet as sdn
+import bfasst
```
dallinjdahl, line 2: is this import used?

scripts/bfasst/compare_waveforms/File_Generation/testbench_generator.py
```diff
@@ -0,0 +1,141 @@
+from re import I
```
dallinjdahl, line 1: is this import used?

scripts/bfasst/compare_waveforms/File_Parsing/parse_diff.py
```diff
@@ -0,0 +1,177 @@
...
+    # Because we don't need the bit_length, it will be discarded. Each symbol and signal are stored in individual arrays and
+    # each entry will correspond with each other. (Note: Could be replaced with a map in the future.)
+    with paths["vcd"][0].open("r") as file:
+        for line in file:
+            if "$var wire" in line:
+                word = ""
+                for i in line:
+                    if i == " ":
+                        newWord = True
+                    if newWord == False:
```
dallinjdahl, line 65: `is False`

scripts/bfasst/compare_waveforms/File_Parsing/parse_diff.py
```diff
@@ -0,0 +1,177 @@
...
+                                    num = int(num)
+                                    num = int(num / 1000)
+                                    line = " " + str(num) + " ns\n"
+                                    isParsed = True
+                    elif symbol == "$":
+                        if "$scope" not in line:
+                            if "$var wire" not in line:
+                                if "$timescale" not in line:
+                                    if "$end" not in line:
+                                        if isParsed == False:
```
dallinjdahl, line 141: `is False`

scripts/bfasst/compare_waveforms/Interface/waveform_interface.py
```diff
@@ -0,0 +1,65 @@
+import yaml
+
+"""A function to check if there are multiple verilog files in a design or not. Used later in parsing stages due to different logic being needed."""
+
+
+def check_multiple_files(design):
+    with open(design.yaml_path) as fp:
```
dallinjdahl, line 7: should specify file open mode (r, w, rw, etc)

scripts/bfasst/compare_waveforms/Interface/waveform_interface.py
```diff
@@ -0,0 +1,65 @@
...
+        design_props = yaml.safe_load(fp)
+
+    multiple_files = False
+
+    for k, v in design_props.items():
```
dallinjdahl, line 12: Since you never use `v`, it would be better to use `for k in design_props.keys():`

scripts/bfasst/compare_waveforms/Interface/waveform_interface.py
```diff
@@ -0,0 +1,65 @@
...
+            "Design has already been tested and came back unequivalent. View Waveforms? Input 1 for yes, 0 for no."
+        )
+        if cont == "0":
+            cont = input(
+                "Ok. Do you want to generate new testbenches and check equivalence again? Input 1 for yes, 0 for no."
+            )
+            if cont == "0":
+                print("Ok. Ending Tests.")
+                return 0
+            else:
```
dallinjdahl, line 45: unncessary else

scripts/bfasst/compare_waveforms/Interface/waveform_interface.py
```diff
@@ -0,0 +1,65 @@
...
+            "Design has already been tested and was equivalent. View Waveforms? Input 1 for yes, 0 for no."
+        )
+        if cont == "0":
+            cont = input(
+                "Ok. Do you want to generate new testbenches and check equivalence again? Input 1 for yes, 0 for no."
+            )
+            if cont == "0":
+                print("Ok. Ending Tests.")
+                return 3
+            else:
```
dallinjdahl, line 60: uneccesary else

scripts/bfasst/compare_waveforms/Templates/get_paths.py
```diff
@@ -0,0 +1,42 @@
+import bfasst
+from bfasst.compare_waveforms.Templates import structs
+
+paths = structs.paths
+
+def get_paths(self, design):
```
dallinjdahl, line 6: get_paths isn't part of a class, so it shouldn't require self.  If it does, it should be part of a class.

scripts/bfasst/compare_waveforms/Templates/structs.py
```diff
@@ -0,0 +1,35 @@
+"""A data structure used to store the list of inputs, outputs, and their respective bits."""
```
dallinjdahl, line 1: Note that you don't need to pre-declare data structures, you should be able to access them as necessary. see `setdefault()`

scripts/bfasst/compare_waveforms/Tools/analyze_graph.py
```diff
@@ -24,44 +26,48 @@ def analyze_graphs(path, module):
...
+    run_vivado = (
+        bfasst.paths.ROOT_PATH / "scripts/bfasst/compare_waveforms/run_vivado.py"
+    )
     base_path = bfasst.paths.ROOT_PATH / "scripts/bfasst/compare_waveforms"
 
     gtkwave = Path(".gtkwaverc")
-    if(gtkwave.exists()):
+    if gtkwave.exists():
         gtkwave.unlink()
     with gtkwave.open("x") as wavefile:
```
dallinjdahl, line 37: r,w,rw?

scripts/bfasst/compare_waveforms/Tools/analyze_graph.py
```diff
@@ -97,21 +112,22 @@ def analyze_graphs(path, module):
...
 
+"""A function primarily meant to check the user's monitor resolution so gtkwave can launch in full-screen mode."""
+
 
-'''A function primarily meant to check the user's monitor resolution so gtkwave can launch in full-screen mode.'''
 def find_resolution():
     output = subprocess.check_output("xrandr")
     output = output.decode()
     temp = Path("temp.txt")
     with temp.open("x") as file:
```
dallinjdahl, line 128: r,w,rw?

scripts/bfasst/compare_waveforms/Tools/analyze_graph.py
```diff
@@ -120,20 +136,20 @@ def find_resolution():
             if "primary" in line:
                 isPrimary = True
             if isPrimary:
-                #The user-resolution is indicated by the * character in a line.
+                # The user-resolution is indicated by the * character in a line.
                 if "*" in line:
                     if foundDisplay == False:
```
dallinjdahl, line 141: `is False`

scripts/bfasst/compare_waveforms/Tools/analyze_graph.py
```diff
@@ -120,20 +136,20 @@ def find_resolution():
...
+                        x = line[0 : line.index("x")]
+                        y = line[line.index("x") + 1 : line.index(" ")]
                         foundDisplay = True
                         temp.unlink()
-                        return(x, y)
-    #If, for whatever reason, the screen resolution can't be found, it defaults to the absolute lowest screen resolution.
-    if(foundDisplay == False):
+                        return (x, y)
+    # If, for whatever reason, the screen resolution can't be found, it defaults to the absolute lowest screen resolution.
+    if foundDisplay == False:
```
dallinjdahl, line 151: `is False`

scripts/bfasst/compare_waveforms/Tools/run_vivado.py
```diff
@@ -0,0 +1,77 @@
...
+base_path_r = sys.argv[4]
+tb_path_r = sys.argv[5]
+tb_name_r = sys.argv[6]
+
+vivado = sys.argv[7]
+base = sys.argv[8]
+
+assert vivado is not None, "VIVADO_PATH environmental variable was not set!"
+
+template = Path(base) / str("template.tcl")
```
dallinjdahl, line 18: Isn't template.tcl already a string?

dallinjdahl (changes requested):
Can you run this through black for formatting?  There are some long lines and odd if formats.  Also, directory names should be lowercase for consistency.  Finally, in the future, pull requests should be limited to a single feature change, with an associated branch.  Then feedback can be much more targeted, and things get processed faster.

JakeEdvenson:
@dallinjdahl Interesting, I ran all of this through black before I committed it, I'm betting it's because I made some last minute changes during the meeting and didn't run it through black again. When you say directory names, are you referring to LOG_FILE_NAME and TOOL_WORK_DIR? Both of those are uppercase because they are uppercase in conformal.py and yosys.py. If not, what in particular isn't lowercase? As for PRs in the future, I can make sure to do that.

scripts/bfasst/compare/waveform.py
JakeEdvenson: Like adding default=CompareStatus.NOT_EQUIVALENT to the function? Out of curiosity, why? The function check_diff always returns true or false. Should I replace that to be true or default? Or should the default be added to run_test?

scripts/bfasst/compare/waveform.py
JakeEdvenson: This is the solution I came to from my understanding of the comment: The only aspect of paths that requires self is the actual build directory. I could just pass in the directory itself like get_paths(self.work_dir, design) and then it creates all of the paths from there. If I'm incorrect on this, I may have misunderstood the comment.

scripts/bfasst/compare_waveforms/File_Generation/waveform_generator.py
JakeEdvenson: This would have been a life-saver when developing this script!!! Once I finish clearing up this PR, I'm going to take a look to see what that changes with the rest of the logic of the tool.

scripts/bfasst/compare_waveforms/Templates/get_paths.py
JakeEdvenson: I think I mentioned this in another comment, but if passing in self.work_dir would work, that would probably be the easiest solutions since it's the only actual usage of self in this file.

scripts/bfasst/compare/waveform.py
dallinjdahl: Sorry, I was talking about check_compare_status.  Currently all the returns are inside if statements, and if none of them are true, it's going to return None, which is inconsistent with the rest of the branches.

scripts/bfasst/compare_waveforms/File_Generation/testbench_generator.py
```diff
@@ -1,68 +1,66 @@
-from re import I
 import numpy as np
 
 
 """A simple function to return the amount of data stored in the input list."""
```
dallinjdahl, line 4: You shouldn't need this line anymore, right?

scripts/bfasst/compare_waveforms/Templates/get_paths.py
```diff
@@ -1,42 +1,84 @@
...
-    )
-    paths["tcl"].append(paths["build_dir"] / (paths["modules"][1] + ".tcl"))
-    paths["tcl"].append(paths["build_dir"] / (paths["modules"][2] + ".tcl"))
-    paths["fst"].append(paths["build_dir"] / (paths["modules"][1] + "_temp.vcd.fst"))
-    paths["fst"].append(paths["build_dir"] / (paths["modules"][2] + "_temp.vcd.fst"))
-    paths["parsed_diff"] = self.work_dir / ("parsed_diff.txt")
-    paths["test"] = paths["build_dir"] / "test.v"
+def get_paths(work_dir, root_path, design):
+    # The base directory that files are stored in
+    paths.setdefault("build_dir", work_dir)
```
dallinjdahl, line 6: Sorry, I was a little unclear.  setdefault is used when accessing a dictionary item, to say get the item if it exists, otherwise return this value.  It doesn't change anything inherently in the dictionary.  I only suggested it because it seemed that you were attempting to avoid appending to a Null value, in the case of a list.  For example, instead of doing something like this:
```python
a = {"paths": []}
a["paths"].append("/usr/local/bin")
```
You can do this:
```python
a = {}
a.setdefault("paths", []).append("/usr/local/bin")
```
and you don't have to worry about appending to null.

scripts/bfasst/compare_waveforms/Templates/structs.py
```diff
@@ -1,5 +1,5 @@
-"""A data structure used to store the list of inputs, outputs, and their respective bits."""
 data = {
+    """A data structure used to store the list of inputs, outputs, and their respective bits."""
     "input_list": [],
```
dallinjdahl, line 3: You might consider using the setdefault approach here as well.

scripts/bfasst/compare_waveforms/File_Generation/waveform_generator.py
```diff
@@ -24,14 +21,12 @@ def generate_VCD(paths, i):
     gtkwave(paths, i)
 
 
-"""Creates a waveform that only takes into account the inputs/outputs (totally ignores all internal components) and then
-takes the input/output graph and turns it into a VCD file to be compared against."""
-
-
 def gtkwave(paths, i):
```
dallinjdahl, line 24: Given that you switched the depth of the vcd generation from all signals to only the top level,  does gtkwave still help?

scripts/bfasst/compare_waveforms/File_Parsing/parse_diff.py
```diff
@@ -28,7 +27,7 @@ def check_diff(paths):
         print("NOT EQUIVALENT! SEE " + str(paths["parsed_diff"]) + " for more info")
```
dallinjdahl, line 27: f_strings

scripts/bfasst/compare_waveforms/File_Generation/tcl_generator.py
```diff
@@ -1,15 +1,13 @@
-"""Generates the first TCL that will be used in gtkwave to create a VCD output."""
-
-
 def generate_first_TCL(paths, data, i):
+    """Generates the first TCL that will be used in gtkwave to create a VCD output."""
     path = paths["build_dir"] / (paths["modules"][i + 1] + ".tcl")
```
dallinjdahl, line 3: f-strings

scripts/bfasst/compare/waveform.py
```diff
@@ -74,41 +74,38 @@ def compare_netlists(self, design, runInterface):
...
                 if paths["tb"][i].exists():
                     paths["tb"][i].unlink()
             self.generate_files(multiple_files)
             if self.run_test():
                 return self.success_status
-            else:
-                return Status(CompareStatus.NOT_EQUIVALENT)
-        elif (
+            return Status(CompareStatus.NOT_EQUIVALENT)
+        if (
```
dallinjdahl, line 86: Unnecessary parens

scripts/bfasst/compare_waveforms/Templates/get_paths.py
dallinjdahl: Although, in this case, it might be easier to just change this:
```python
    paths["path"].append(design.impl_netlist_path)
    paths["path"].append(design.reversed_netlist_path)
```
to this:
```python
    paths["path"] = []
    paths["path"].append(design.impl_netlist_path)
    paths["path"].append(design.reversed_netlist_path)
```
That's probably the cleanest to me.

dallinjdahl (changes requested):
Looks much better.  Just a couple more comments.  Regarding the directory names, I was referring to the Tools, Interface, Templates, etc. directories.  Everything else in the project is lowercase, so it would be more consistent to name them tools, interface, templates, etc.  Also, just for future reference, the difference between `==` and `is` is that `==` compares the contents of each object, while `is` checks to see if they are the same object (basically a pointer comparison if that helps).  Since `True` and `False` are singleton objects (there is only one instance of them ever), `is` is much more efficient.

scripts/bfasst/compare/waveform.py
dallinjdahl: That approach works.  My initial thought was to bring it back into the class, but I think I like this better.  You could also avoid creating paths separately by just creating it in get_paths, and then returning it.  Anyway, this architecture is much better :)

scripts/bfasst/compare_waveforms/Templates/structs.py
dallinjdahl: The setdefault would probably go in parse_files.py

scripts/bfasst/compare/waveform.py
dallinjdahl: Sorry, the if statement has unecessary parens.  Just so you're aware, the comments show up just below the relevant line, so I'm doing my best to put the comments right underneath the applicable section.

scripts/bfasst/compare_waveforms/file_parsing/parse_diff.py
```diff
@@ -8,7 +8,7 @@ def check_diff(paths):
 
     # Finds how many lines are different in the two files.
     dif = subprocess.getoutput(
-        ["diff -c " + str(paths["vcd"][0]) + " " + str(paths["vcd"][1])]
+        [f"diff -c {str(paths['vcd'][0])} {str(paths['vcd'][1])}"]
```
dallinjdahl, line 11: str is unnecessary when using f-strings.

scripts/bfasst/compare_waveforms/file_parsing/parse_files.py
```diff
@@ -58,19 +73,21 @@ def parse_multiple(file, reversed_file):
...
+                    total_list.append(port.name)
+                    input_bits_list.append(len(port.pins) - 1)
+            if "clk" in input_list:
+                if input_list.index("clk") == 0:
+                    input_list.append(input_list.pop(0))
         else:
             not_port = False
+    data = set_data(
+        input_list, output_list, input_bits_list, output_bits_list, total_list
+    )
```
dallinjdahl, line 90: Do you know you are going to have data in all sub_elements of `data`, ie input_list, output_list, input_bits_list, output_bits_list, total_list?  If that's the case, then setdefault won't do anything helpful.  Simply setting up `data["input"] = []` before accessing it is probably faster and cleaner.

The reason you would use setdefault is if you want to make sure you don't overwrite anything, but if nothing is there, you want to make the value for that key a list (or string, or number, etc.).  If you may or may not have a dictionary item initialized at the end of the function, then it makes sense to use setdefault when you need to, because that preserves the non-initialized state if you never use it, but gives you a deterministic default when you do use it.  That means that you never have empty lists in a dictionary.  If the key exists in the dictionary, it's value is guaranteed to be non-empty, since it's only empty as a default right before you change it.  Does that make sense?

scripts/bfasst/compare_waveforms/templates/get_paths.py
```diff
@@ -0,0 +1,58 @@
+paths = {}
```
dallinjdahl, line 1: This should probably be inside `get_paths`.

dallinjdahl (changes requested):
Sorry my explanation of setdefault was so confusing.  I tried to clear it up a little.  Let me know if you have further questions.

scripts/bfasst/compare/waveform.py
```diff
@@ -73,539 +49,113 @@ def compare_netlists(self, design):
...
-                    self.generate_files(design, giveWarning)
-                    if self.run_test(design):
-                        return self.success_status
-                    else:
-                        return Status(CompareStatus.NOT_EQUIVALENT)
-            else:
-                analyze_graph.analyze_graphs(build_dir, module)
-                return self.success_status
+        if (
+            runInterface
```
dallinjdahl, line 57: unnecessary parens

scripts/bfasst/compare/waveform.py
```diff
@@ -73,539 +49,113 @@ def compare_netlists(self, design):
...
-                        newFile.write(line)
-                    elif "input" in line:
-                        newFile.write(line)
-                    elif "output" in line:
-                        newFile.write(line)
+                    file_rewriter.fix_file(
+                        paths, i
+                    )  # Rewrites the files to have correct module names
+                    if (
+                        multiple_files
```
dallinjdahl, line 119: unnecessary parens

scripts/bfasst/compare/waveform.py
```diff
@@ -73,539 +49,113 @@ def compare_netlists(self, design):
...
-                    data["random_list"].append(
-                        np.random.randint(low=0, high=2, size=int(test_num))
-                    )
-                else:
-                    data["random_list"].append(
-                        np.random.randint(
-                            low=0, high=(2 ** (int(bits) + 1) - 1), size=int(test_num)
-                        )
+                if (
+                    i == 0
```
dallinjdahl, line 128: unnecessary parens

scripts/bfasst/compare_waveforms/templates/get_paths.py
```diff
@@ -0,0 +1,56 @@
+def get_paths(work_dir, root_path, design):
+    paths = {}
+    # The base directory that files are stored in
+    paths.setdefault("build_dir", work_dir)
```
dallinjdahl, line 4: Since paths is empty, setdefault does nothing here.

scripts/bfasst/compare_waveforms/templates/get_paths.py
```diff
@@ -0,0 +1,56 @@
...
+    # Cells sim is used as a reference for any modules that are defined such as IBUF
+    paths.setdefault(
```
dallinjdahl, line 6: likewise

scripts/bfasst/compare_waveforms/templates/get_paths.py
```diff
@@ -0,0 +1,56 @@
...
+        "cells_sim", root_path / ("third_party/yosys/techlibs/xilinx/cells_sim.v")
+    )
+    # Path to the sample testbench used for creating the automatic testbench
+    paths.setdefault(
```
dallinjdahl, line 10: likewise

dallinjdahl (changes requested):
Looking good.  A couple of things that I think you missed from my last set of comments.  Also, this needs to be rebased on top of the master branch to allow me to pull it in.

dallinjdahl (approved):
Looks great!

PR #63, opened 2022-09-07
Add Exception Handling

reillymck:


jgoeders:
@reillymck Looks good.  Have you verified that this still works with run_experiment.py?

reillymck:
Update I see that the parallel threads are all killed if an exception is raised. I will add a fix so that if the BfasstException is raised that it is just attached to the status list instead of killing all the threads.

reillymck:
@jgoeders I think I have made the proper adjustments in the code for the multithreading. However, I am trying to test it with one of the .yaml in the experiments folder. Due to code errors in the current main branch (for instance my current open issue), some of the .yaml files do not run correctly. Is there a proper .yaml for the CI test? @dallinjdahl which scripts are you planning on using for the CI test? I am not sure if just running the code under "check installation" is enough since we have multithreading + we have the experiment .yaml files. Maybe the CI should test that all experiment .yaml's still work with new PRs?

reillymck:
These features will be added in the flow_args pull request.

PR #69, opened 2022-09-17
Sam designs

TheDHCreator:
I think this was changed in another PR already but if so, just let me know and close the request. I was just pulling in everyone's recent changes. Thanks. - Sam


reillymck:
So environments are activated using . ./ instead of source and python3 should then link to the active environment. So there is no need to copy bin into the environment. Let us know if you encounter any problems with running it.

PR #70, opened 2022-09-20
Flow args

reillymck:
This pull request adds the following features:
- Error status triggers an exception. However, there is a parameter in the status constructor to prevent the exception from being raised. Checking for error status is now done in a try-catch block if you do not want the function to fail/return. 
- The framework for flow arguments is implemented. Xilinx flow arguments are supported for synthesis (SYNTH), including overwriting the default arguments for -dsp and -top. Xilinx implementation has been edited to work with designs synthesized out of context (constraints are not loaded and bitstream generation is skipped). CMP supports the argument "xilinx" or "lattice" (not case sensitive). This is directly fed into the VENDOR parameter when applicable.
    *Flow arguments include SYNTH, IMPL, MAP, EQUIV, CMP, & ERR
- Flow args are parsed into a dictionary with keys being the args and a string being the value. Currently for xilinx synthesis the entire SYNTH string is appended to the synth_design command. The flow args incorporate the new usage:
```
--synth FLOW_ARGS     Arguments for synthesis
--impl FLOW_ARGS      Arguments for implementation
--map FLOW_ARGS       Arguments for mapping
--equiv FLOW_ARGS     Arguments for equivalence
--cmp FLOW_ARGS       Arguments for comparison
--err FLOW_ARGS       Arguements for error flow
```
An example for xilinx flow would be:
```
run_design.py design_dir xilinx --synth "-mode out_of_context -flatten_hierarchy"
```
Or in the experiment .yaml file include:
```
synth: "-mode out_of_context -flatten_hierarchy"
```
- The flows.py file has been updated to include wrapper functions for each flow piece, ie: xilinx synthesis, yosys synthesis, yosys compare, etc. The flow functions now reference these wrappers. All flow functions accept a flow_args argument and all wrapper functions also accept the argument. The tool constructor also includes an optional flow_args parameter.
- Flow arguments can be specified in experiment .yaml files. They are not case-sensitive.
- run_experiment.py prints the output of each thread to build_dir/experiment_name/design/design_name.log
    *The running job and job completion status is still printed to stdout
    *References to print_to_stdout have been removed.
- run_experiment still only crashes when a Python Exception is raised and is not effected by the BfasstException (Status exceptions).

I have added a new .yaml "xilinx_ooc.yaml" which I used to test these new features. The command I used is
```
python scripts/run_experiment.py experiments/xilinx_ooc.yaml -j 8
```

### DIFF
--- a/experiments/xilinx_ooc.yaml
+++ b/experiments/xilinx_ooc.yaml
@@ -1,0 +1,6 @@
+designs:
+  - ooc/
+
+flow: xilinx
+
+synth: "-mode out_of_context"
--- a/scripts/bfasst/compare/base.py
+++ b/scripts/bfasst/compare/base.py
@@ -8,8 +8,8 @@ from bfasst.utils import print_color
 
 
 class CompareTool(Tool):
-    def __init__(self, cwd) -> None:
-        super().__init__(cwd)
+    def __init__(self, cwd, flow_args="") -> None:
+        super().__init__(cwd, flow_args)
         self.success_status = Status(CompareStatus.SUCCESS)
 
     # This method should run netlist comparison.  It should return
--- a/scripts/bfasst/compare/conformal.py
+++ b/scripts/bfasst/compare/conformal.py
@@ -33,9 +33,7 @@ class Conformal_CompareTool(CompareTool):
         assert type(vendor) is flows.Vendor
         self.vendor = vendor
 
-    def compare_netlists(self, design, mapping_algorithm, print_to_stdout=True):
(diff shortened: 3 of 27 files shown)


jgoeders:
@reillymck

reillymck:
@jgoeders I guess none. Pablo added EQUIV, but I added CMP. I did not realize they overlapped.

reillymck:
Right now the CMP Vendor arg is only fed into Conformal_CompareTool

dallinjdahl (approved):
Looks good

PR #73, opened 2022-09-22
Pylint on pr

reillymck:
Run pylint (for errors only) on all new pull requests. This pr also includes fixes to all previous errors in the code base.


dallinjdahl:
Do we want to disable warnings?  I think that would be really helpful...

dallinjdahl:
Also the R set was helpful to catch things like unecessary else...

reillymck:
@dallinjdahl I do think warnings should be included. However, I do think we can disable some warnings, like  "W1514: Using open without explicitly specifying an encoding." I just set this up to do errors only, so that we can have something in place while we get the warnings figured out. Dr. Goeders gave me a config file that he uses to control pylint, and I'll be looking over it. I figured we could add this now, to stop any pr with code that just doesn't work from being merged.

reillymck:
I agree that the R set can also be helpful. Between W & R, we have almost 1k lines of output of stuff to be fixed according to pylint. That will take some time to analyze, so again, this error check is just something to get us started while we fix up everything else. As you can see, passing the error checking alone required editing 33 files.

dallinjdahl (approved):
Looks good!

PR #82, opened 2022-10-10
Improved difference checking algorithm

JakeEdvenson:
A lot of false flags were being raised when I was simply running a diff on the two VCD files, so I changed the whole system so now it just analyzes the state of every signal at every time and then compares the two files and checks if they are the same. It's more reliable and easier to read from a user perspective.


JakeEdvenson:
@dallinjdahl 
Any ideas on this import error from Pylint? It works in the environment I run on my computer.
`scripts/bfasst/compare_waveforms/tools/analyze_graph.py:5:0: E0401: Unable to import 'bfasst' (import-error)`

dallinjdahl (changes requested):
It looks like you still need to migrate to argparse.  It also looks like most of the issues you're having with the coding standard are from ad-hoc parsers.  You might want to look at the following to have some ideas about how to do it more cleanly:
https://en.wikipedia.org/wiki/Recursive_descent_parser
https://en.wikipedia.org/wiki/Shift-reduce_parser

Here is a python package that might help:
https://github.com/r1chardj0n3s/parse

You could also look at map/filter, which might make it easier:
https://stackabuse.com/map-filter-and-reduce-in-python-with-examples/

Or list comprehensions:
https://stackabuse.com/list-comprehensions-in-python/

Let me know if you have any questions.  I'll be adding some resources to the coding standard to make it easier to follow.

dallinjdahl (changes requested):
It's looking much better.  If you could refactor the main function in run_vivado.py so it passes the parsed arguments to another function to do the work, then people importing run_vivado can use that workflow programmatically instead of having to invoke the script as a separate process.

scripts/bfasst/compare_waveforms/tools/run_vivado.py
```diff
@@ -1,15 +1,27 @@
...
+
+    parser = argparse.ArgumentParser(description="Launch Vivado.")
+    parser.add_argument("base", metavar="Base", help="Base path to files.")
+    parser.add_argument("module", metavar="Module", help="Module name.")
+    parser.add_argument("temp", metavar="Temp Folder", help=
+    "Temporary Location for TCL files.")
+    parser.add_argument("vivado", metavar="Vivado", help="Location of Vivado Launcher.")
+
+    global args
+    args = parser.parse_args()
```
dallinjdahl: Just a thought, if you `return parser.parse_args()`, then you don't need the global, you can receive the result in main.

scripts/bfasst/compare_waveforms/tools/run_vivado.py
```diff
@@ -47,9 +49,9 @@ def create_tcl(template, temp_tcl, args):
                 output.write(line)
 
 
-def main():
+def launch_vivado():
```
dallinjdahl, line 52: `launch_vivado` should take `base`, `module`, `temp`, and `vivado` as parameters to the function, and be passed those by main.  That way you can call `launch_vivado` from another script.  Currently you can't, because `args` will be `None`.

scripts/bfasst/compare_waveforms/file_parsing/parse_diff.py
```diff
@@ -21,39 +26,46 @@ def parse_io(line, input_output, words, new_word):
...
+
         if i == " ":
             new_word = True
+
         if new_word is False:
             word = word + i
+
         else:
+
             if (word != "$var") & (word != input_output):
```
dallinjdahl, line 40: I'm pretty sure this ampersand is a bitwise operator in python, you probably want `and`.

scripts/bfasst/compare_waveforms/file_parsing/parse_diff.py
```diff
@@ -1,18 +1,23 @@
...
+    data["state"] = []
+    data["binary_state"] = []
+    data["returned_signal"] = []
+    data["past_header"] = False
+    data["past_definitions"] = False
+
+    return data
 
 
 def parse_io(line, input_output, words, new_word):
```
dallinjdahl, line 23: Just fyi, as long as your code works and follows the coding standard, I'll approve it. However, this function in particular is very amenable to some of the documented parser techniques I listed in the examples section of the coding standard.  I don't know if you've taken cs 236, but those techniques are exactly for this purpose.  You also might look at python's `split` function, which looks like it would greatly reduce your logic.

dallinjdahl (changes requested):
Almost there.  Mostly just the main function, and some rework of the ampersand.  There is some additional feedback, but mostly just suggestions.

dallinjdahl (approved):
Looks great!

PR #93, opened 2022-10-24
Fix to Issue#92

JakeEdvenson:
I opened an issue bfasst not handing system verilog files, I tracked down the issue to this line of code. This edit seems to fix the issue by getting the only top file from the filtered list rather than by assuming that the list will always return a verilog file. Feel free to suggest a better way to do this if there is one, but I figured this should make fixing the issue a lot easier.

### DIFF
--- a/scripts/bfasst/design.py
+++ b/scripts/bfasst/design.py
@@ -155,7 +155,7 @@ class Design:
                 self.path,
             )
 
-        return top_paths[0]
+        return list(valid_paths)[0]
 
     def hdl_by_suffix(self, *suffixes):
         return (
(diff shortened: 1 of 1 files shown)


scripts/bfasst/design.py
```diff
@@ -155,7 +155,7 @@ def find_top_file(self):
                 self.path,
             )
 
-        return top_paths[0]
+        return list(valid_paths)[0]
```
dallinjdahl, line 158: try `next(valid_paths)` instead

dallinjdahl (changes requested):
Good catch!  To avoid creating a whole list when we only use the first item, I suggest using the next function.

dallinjdahl (approved):
Looks great!

PR #97, opened 2022-10-26
Waveform tool updated to coding standard

JakeEdvenson:
I believe this should catch everything that was violating the coding standard (at least according to the pylint extension of vscode it should.) I also included pylint into the settings.json file in .vscode so that it will default to using it on everyones Vscode.

### DIFF
--- a/scripts/bfasst/compare/waveform.py
+++ b/scripts/bfasst/compare/waveform.py
@@ -91,7 +91,7 @@ class WaveformCompareTool(CompareTool):
         it calls the testbench generators, finally it calls the TCL generators. It then increments
         to the next file and clears the data structure."""
         test_num = 100  # Change this number if you want to run more or less than 100 tests in the
-        #testbench.
+        # testbench.
         data = {}  # Contains all of the IOs for the design.
         file_rewriter.copy_files(paths)
         # Creates copies of the netlists that will be modified by the file_rewriter
@@ -110,7 +110,7 @@ class WaveformCompareTool(CompareTool):
                     # Rewrites the files to have correct module names
                     if multiple_files:
                         # The logic for how to parse the file depends on whether or not there are
-                        #multiple verilog files involved in a design
+                        # multiple verilog files involved in a design
                         data = parse_files.parse_reversed(paths, i)
                         # Finds the IO names and bit sizes
                     else:
--- a/scripts/bfasst/compare/waveform.py
+++ b/scripts/bfasst/compare/waveform.py
@@ -1,9 +1,12 @@
-# Waveform equivalence checker. Designed by Jake Edvenson.
-# Relies on gtkwave, icarus-verilog, spydrnet, and numpy.
-# To use this script, use ./scripts/run_design.py (DESIGN_PATH) xilinx_yosys_waveform
-# The rest of the script should be self-explanitory. I've included a lot of user-prompts to make it very user-friendly.
-# If vivado or F4PGA ever change there netlist-generating style, this file may need to be edited.
-# I would check the fix_file function because this file currently alters the netlists so that spydrnet can parse them.
+"""Waveform equivalence checker. Designed by Jake Edvenson.
+    Relies on gtkwave, icarus-verilog, spydrnet, and numpy.
+    To use this script, use ./scripts/run_design.py (DESIGN_PATH)
+    xilinx_yosys_waveform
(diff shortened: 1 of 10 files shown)


scripts/bfasst/compare/waveform.py
```diff
@@ -144,12 +156,12 @@ def check_compare_status(self, log_path, default=CompareStatus.NOT_EQUIVALENT):
             return Status(CompareStatus.TIMEOUT)
 
         # Regex search for result
-        m = re.search(r"Equivalence successfully proven!", log_text, re.M)
-        if m:
+        i = re.search(r"Equivalence successfully proven!", log_text, re.M)
```
dallinjdahl, line 159: This doesn't need to be a raw string.  More readable to just have it be a normal string.

scripts/bfasst/compare/waveform.py
```diff
@@ -144,12 +156,12 @@ def check_compare_status(self, log_path, default=CompareStatus.NOT_EQUIVALENT):
...
+        if i:
             return Status(CompareStatus.SUCCESS)
 
-        m = re.search(r"ERROR", log_text, re.M)
-        if m:
+        i = re.search(r"ERROR", log_text, re.M)
```
dallinjdahl, line 163: same as above.

scripts/bfasst/compare_waveforms/file_generation/testbench_generator.py
```diff
@@ -1,110 +1,187 @@
...
     return len(data["input_list"]) + len(data["output_list"])
 
 
+def generate_random(data, random_list, test_num):
+
+    """Generates the random signals for the testbench"""
+
+    for bits in data["input_bits_list"]:
+        if bits == 0:
+            random_list.append(np.random.randint(low=0, high=2, size=int(test_num)))
```
dallinjdahl, line 27: Is there a reason you use numpy over the standard random package?

scripts/bfasst/compare_waveforms/file_generation/testbench_generator.py
```diff
@@ -1,110 +1,187 @@
...
+
+
 def generate_first_testbench(paths, test_num, data, i):
-    """This function creates the initial testbench that will be modified by the reversed-netlist. It reads in a sample testbench
-    and replaces certain variables with the corresponding information from the data structure. It also sets the variables equal
+
+    """This function creates the initial testbench that will be modified by the reversed-netlist.
+    It reads in a sample testbench
+    and replaces certain variables with the corresponding information from the data structure. It
+    also sets the variables equal
```
dallinjdahl, line 168: You could join these lines to make it flow better.  On linux, you can run this block through `fmt` and it will spit out a nicely formatted version.

scripts/bfasst/compare_waveforms/file_parsing/parse_files.py
```diff
@@ -1,8 +1,29 @@
...
+
 def parse(file):
-    """Uses spydrnet to analyze the netlist and add the names of all inputs, outputs, and their respective bit sizes to the data structure."""
+
+    """Uses spydrnet to analyze the netlist and add the names of all inputs, outputs, and their
+    respective bit sizes to the data structure."""
+
     netlist = sdn.parse(str(file))
     library = netlist.libraries[0]
     definition = library.definitions[0]
```
dallinjdahl, line 29: Consider using `netlist.top_instance().parent()` to get the definition.  This has the added benefit of working with edif files.

dallinjdahl (changes requested):
Looks good.  Mostly just fix the raw strings and the comment formatting.

scripts/bfasst/compare/waveform.py
JakeEdvenson: This check_compare_status function is from whoever made conformal.py, onespin.py, and yosys.py. I'll make sure to fix it, but I suspect these changes will have to be done on the rest of the python scripts in compare as well.

scripts/bfasst/compare_waveforms/file_parsing/parse_files.py
JakeEdvenson: I'm running some tests on this function and currently it's telling me that parent definition is undefined. Reference definition seems to be defined, but I'm not sure what that really does. I'll see if the documentation can clear that up.

dallinjdahl (approved):
Looks great!

PR #99, opened 2022-11-03
Change pylint to run error checking on all files

reillymck:
Also:
- Docstrings for functions > 10 lines instead of >=.
- Acceptable variable name list expanded.
- Removed allowance for variables named f[0-9]

Fixes were made to pass pylint error checking.


dallinjdahl (approved):
Looks great (except for clk and val :))

PR #100, opened 2022-11-14
Removed dependencies on bfasst

JakeEdvenson:
I'm preparing the tool to be put in a separate directory as a third_party tool rather than bfasst exclusive. Currently, it works as it's own folder, so these changes are a step in that direction.


reillymck:
Please do not request user input, use function parameters and command line arguments. I noticed this in a few files. Thanks.

scripts/bfasst/compare_waveforms/compare_waveforms.py
```diff
@@ -68,44 +72,73 @@ def run_test(paths):
...
 
     parser.add_argument(
-        "-q", "--quick", action='store_true',
-        help="Skips all interface interactions. Specify tests or they default to 100.",
-        default = False)
+        "-t",
+        "--tests",
+        action="store",
+        help="The number of tests to run. If not set, defaults to 100.",
+        default=0,
```
reillymck, line 126: Set the default to 100 here.

reillymck:
Good addition with the tests. Please incorporate my previous comment, and then I will merge.

JakeEdvenson:
@reillymck I might be missing something, but I believe I did incorporate your last comment. Default should be set to 100 now and user input shouldn't be requested anymore.

reillymck:
I left a comment directly on waveform.py

JakeEdvenson:
@reillymck I've looked through the waveform.py file and all previous commits with it, I'm not seeing any comments. The only comment I've seen directly on a file in this PR is this one here (which should be resolved):
image

On the main page, I only see your comment about user input, the direct comment on compare_waveform.py, your comment 18 minutes ago, and your response.
Should I be looking somewhere else?

JakeEdvenson:
@reillymck Is there another way to view comments aside from this?
image

scripts/bfasst/compare/waveform.py
```diff
@@ -49,104 +45,76 @@ def compare_netlists(self, design, runInterface):
...
+
+        if design is None:
+            return True
+
+        with open(design.yaml_path, "r") as fp:
+            design_props = yaml.safe_load(fp)
+
+        multiple_files = False
+
+        for k in design_props.keys():
```
reillymck, line 108: Last thing:
```python
return ("included_all_verilog_files" in design_props) or ("include_all_system_verilog_files" in design_props)
```
This should replace lines 106-115

reillymck (changes requested):

reillymck:
Is it visible now?

reillymck (approved):

PR #101, opened 2022-11-30
Move wafove to third party

JakeEdvenson:
Here's my first attempt on getting this set up. Currently, it works during the make process on my computer and now is a part of bfasst's virtual environment. If you want to see what wafove's scripts look like, it can be found here: https://github.com/byuccl/WaFoVe


JakeEdvenson:
@reillymck @dallinjdahl Not sure which one of you wants to check this PR so I just requested both of you.

JakeEdvenson:
@dallinjdahl @reillymck Any chance ya'll could check that this implementation is how I'm supposed to add a third party module to bfasst?

scripts/bfasst/compare/waveform.py
```diff
@@ -79,8 +79,8 @@ def compare_netlists(self, design, run_waveform, tests, vivado):
 
         compare_waveforms.generate_files(multiple_files, paths, tests)
 
-        if compare_waveforms.run_test(paths):
-            Status(CompareStatus.SUCCESS)
+        if compare_waveforms.run_test(paths) is True:
```
dallinjdahl, line 82: You shouldn't need to compare with True in an if statement.  Just use the result of the method invocation as the condition.
```suggestion
        if compare_waveforms.run_test(paths):
```

Makefile
```diff
@@ -1,11 +1,12 @@
 IN_ENV = if [ -e .venv/bin/activate ]; then . .venv/bin/activate; fi;
 
 
-install: packages venv python_packages install_fasm2bels install_yosys
+install: packages install_wafove venv python_packages install_fasm2bels install_yosys
 
 venv:
 	python3 -m venv .venv
 	$(IN_ENV) python3 -m pip install -U pip
+	$(IN_ENV) python3 -m pip install -e third_party/WaFoVe
```
dallinjdahl, line 9: this is probably the wrong place for `pip install`ing your package.  If you decide that `pip` is the way to go, it should either be added to the `requirements.txt` to be picked up by the `python_packages` target, or the `pip install` should be a part of the `install_wafove` target.

Makefile
```diff
@@ -42,12 +43,17 @@ rapidwright:
 	cd third_party/RapidWright && make
 
 install_fasm2bels:
-	git submodule init
-	git submodule update
+	git submodule init third_party/fasm2bels
+	git submodule update third_party/fasm2bels
```
dallinjdahl, line 47: I know this isn't your code, but since we have multiple submodules we're initializing, we should probably add another target, say `submodules`, that runs `git submodule init` and `git submodule update` for everything.  Also, I'm not as familiar with pip install as @reillymck, but do we need to both `pip install` wafove and run `git pull && make`?  At first glance it seems that one or the other would be sufficient.

Makefile
```diff
@@ -42,12 +43,17 @@ rapidwright:
...
 	cd third_party/fasm2bels && make env
 	$(IN_ENV) cd third_party/fasm2bels && make build
 	cd third_party/fasm2bels && make test-py
 
+install_wafove:
+	cd third_party/WaFoVe && git pull origin master
```
dallinjdahl, line 53: We should not run `git pull origin master`, since one key aspect of submodules is that they are tied to a particular commit.  This enables reproducible builds.  If we want to update bfasst, we can do that separately from updating wafove.
```suggestion

```

Makefile
```diff
@@ -42,12 +43,17 @@ rapidwright:
 	cd third_party/RapidWright && make
```
dallinjdahl, line 44: This is what I'm thinking for the submodule target:
```suggestion

submodules:
	git submodule init
	git submodule update

```
and then add this as a dependency for all the third party install mechanisms.

dallinjdahl (changes requested):
Looks like good work.  Just a couple of changes.  Also, have you tested the install process?  As in, have you deleted your local install of bfasst and attempted to install everything as illustrated in the readme?  If adding wafove requires extra work, you might consider adding that to the readme.

Makefile
JakeEdvenson: I tried an alternate approach where instead I defined a function that would do this same thing since I wasn't able to get submodule targets working like you're looking for.

JakeEdvenson:
@dallinjdahl I believe I fixed your primary issues with the PR, any ideas on how to fix this issue now with the auto-checker?

Makefile
dallinjdahl: The issue with this approach is you will run `git submodule init`  and `git submodule update` once for each install target.  I'll make some suggested changes that should do the trick.

Makefile
```diff
@@ -33,26 +37,22 @@ capnproto_java:
...
+	$(call submodules, third_party/fasm2bels)
 	cd third_party/fasm2bels && make env
 	$(IN_ENV) cd third_party/fasm2bels && make build
 	cd third_party/fasm2bels && make test-py
 
 install_wafove:
-	cd third_party/WaFoVe && git pull origin master
-	cd third_party/WaFoVe && make build
-	cd third_party/WaFoVe && make run_test
+	$(call submodules, third_party/WaFoVe)
```
dallinjdahl, line 54:
```suggestion
```

dallinjdahl (changes requested):
I'm curious as to what didn't work.  If it's rapidwright that caused the issues, that makes sense, and I think we have to treat rapidwright as a special case.  In general, you only have to run `git submodule init && git submodule update` once for all the submodules.  I like your changes to the readme and your python code. As to the pylint stuff, I'll keep looking at it.

Makefile
```diff
@@ -33,26 +37,22 @@ capnproto_java:
...
-install_fasm2bels:
-	git submodule init third_party/fasm2bels
-	git submodule update third_party/fasm2bels
+install_fasm2bels: 
+	$(call submodules, third_party/fasm2bels)
 	cd third_party/fasm2bels && make env
 	$(IN_ENV) cd third_party/fasm2bels && make build
 	cd third_party/fasm2bels && make test-py
 
 install_wafove:
```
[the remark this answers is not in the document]
JakeEdvenson: It is rapidwright that caused the issue. I'll try it the way you've provided and leave rapidwright the way it was.

JakeEdvenson:
@dallinjdahl Ok, I've ran another full installation on this commit and everything is working correctly. I'll look into what's causing the import paths issue in the meantime.

reillymck:
We will have to edit the pylint run to include rapidwright. I'll get that added.

reillymck:
Hey, the imports still fail even if wafove is downloaded. Can you look into this? Please note, I changed the pylint file in your branch, so you need to `git pull` before making changes

JakeEdvenson:
@reillymck So, I just tested this in my local repo with a fresh install and it installs correctly, but pylint apparently is still having import issues.

reillymck:
Please fix the comment I left, and then run this on your branch:
```
git fetch origin
git checkout yourBranch
git reset $(git merge-base origin/main $(git branch --show-current))
git add -A
```
Thanks.

JakeEdvenson:
@reillymck Okay, I put all of the commits together. What comment are you referring to in your message? I'm not seeing one left on any of the files nor am I seeing it in slack.

reillymck:
Sorry I'm realizing that I need to complete reviews for comments to show. Basically, you are trying to change scripts/__init__.py and as you can see in the master branch this is an empty file currently.

reillymck (approved):

PR #106, opened 2023-02-08
Xilinx Physical Netlist

jgoeders:
This is the first step for a Xilinx physical netlist generator tool.  It only handles LUTs at this point, and doesn't yet handle fractured LUTs (2 logical LUTs being mapped to the same physical LUT)


Makefile
```diff
@@ -41,6 +41,10 @@ rapidwright:
 	cd third_party/RapidWright/interchange && make
 	cd third_party/RapidWright && make
 
+# Grab a binary release of rapidwright to use with Python/JPype scripts
+rapidwright_bin:
+	cd third_party && wget https://github.com/Xilinx/RapidWright/releases/download/v2022.2.1-beta/rapidwright-2022.2.1-standalone-lin64.jar
+
```
dallinjdahl, line 47: Should this replace the current rapidwright target?  Should we just use the binaries?

Makefile
```diff
@@ -51,6 +55,7 @@ install_fasm2bels:
 env:
 	echo ". `pwd`/third_party/rapidwright.sh" > "env.sh"
 	echo "export INTERCHANGE_SCHEMA_PATH=`pwd`/third_party/RapidWright/interchange/fpga-interchange-schema/interchange" >> "env.sh"
+	echo "export VIVADO_PATH=/tools/Xilinx/Vivado/2022.2/bin/vivado" >> "env.sh"
```
dallinjdahl, line 58: This is also specified in config.py.  Do we need it both places?

dallinjdahl (commented):
Sorry you had to go reformat everything.  Sam is doing the same right now.  I think the code looks good, I'm just a little wary of duplication in the Makefile?

Makefile
jgoeders: ...I don't know.  The rapidwright we clone has the interchange project within it that we build separately.  I don't know if you get that if you download the binary.  Likewise I don't know how to run rapidwright via Python with the cloned source and not a precompiled jar....

Makefile
jgoeders: Good point.  The environment variable needs to exist for the fasm2bels build, so I'm just fixing that here.

But we could update our code to use this environment variable instead of the one in config.py.  Or update config.py to pull from the environment variable?  What do you think?

dallinjdahl: I think that since the makefile is part of the setup, we can't assume the config.py is up to date, so config.py should pull from the environment.  How does that sound?

Makefile
reillymck: You should just do pip install rapidwright. See my comments below.

requirements.txt
```diff
@@ -8,4 +8,5 @@ pyyaml
 numpy
 spydrnet
 werkzeug
-pylint
\ No newline at end of file
+pylint
+JPype1
```
reillymck, line 12: Maybe also just put the rapidwright package here and remove the rapidwright_bin recipe

scripts/bfasst/transform/xilinx_phys_netlist.py
```diff
@@ -0,0 +1,161 @@
...
+import subprocess
+import jpype
+import jpype.imports
+
+from bfasst.config import VIVADO_BIN_PATH
+from bfasst.paths import THIRD_PARTY_PATH
+from bfasst.status import Status, TransformStatus
+from bfasst.transform.base import TransformTool
+
+jpype.startJVM(classpath=[str(THIRD_PARTY_PATH / "rapidwright-2022.2.1-standalone-lin64.jar")])
```
reillymck, line 14: Can we do:
```
try:
    import rapidwright  # This starts jpype, and works if you have only run `pip install rapidwright`
except ImportError:
    jpype.startJVM(classpath=[str(THIRD_PARTY_PATH / "rapidwright-2022.2.1-standalone-lin64.jar")])
```

reillymck: If we add rapidwright to the requirements.txt, then you can just run `import rapidwright` to start JPype and then you can do your com.xilinx.rapidwright imports. Unless you are trying to use a local, modified version of rapidwright with python?

jgoeders: Any reason for the `try`?  If we are going to install the pip package, shouldn't we just assume it exists?

jgoeders:
@dallinjdahl 
>Sorry you had to go reformat everything.
I'm not quite sure what you mean?  ...I don't think I had to do anything extra.

scripts/bfasst/transform/xilinx_phys_netlist.py
dallinjdahl, line 14:
```suggestion
jpype.startJVM(classpath=[str(THIRD_PARTY_PATH / "rapidwright/bin")])
```
This would allow you to use the class files directly, I think...

dallinjdahl:
> Sorry you had to go reformat everything.

    I'm not quite sure what you mean? ...I don't think I had to do anything extra.

Mostly just renaming things

Makefile
dallinjdahl: if we do pip install rapdwright, do we still get the interchange project?

scripts/bfasst/transform/xilinx_phys_netlist.py
reillymck: No reason for the try in that case. Sorry I wrote this comment first before the other ones. Yes, we can just do the import rapidwright.

jgoeders: I'll give it a try

dallinjdahl: Note that with pip install rapidwright, we effectively just get the jar, so if we want the interchange project as well, I think we'd need to download that separately.

dallinjdahl: Just for reference, this comes from:
https://docs.oracle.com/javase/1.5.0/docs/tooldocs/findingclasses.html
> A class file has a subpath name that reflects the class's fully-qualified name. For example, if the class `com.mypackage.MyClass` is stored under `/myclasses`, then `/myclasses` must be in the user class path and the full path to the class file must be `/myclasses/com/mypackage/MyClass.class`. If the class is stored in an archive named `myclasses.jar`, then `myclasses.jar` must be in the user class path, and the class file must be stored in the archive as `com/mypackage/MyClass.class`.

Makefile
jgoeders: I've updated to use our existing rapidwright code.

Makefile
jgoeders: Fixed.

requirements.txt
jgoeders: I've updated to use our existing rapidwright code.

scripts/bfasst/transform/xilinx_phys_netlist.py
jgoeders: I've updated to use our existing rapidwright code:
```python
jpype.startJVM(
    classpath=[
        str(THIRD_PARTY_PATH / "RapidWright" / "bin"),
        *(str(s) for s in (THIRD_PARTY_PATH / "RapidWright" / "jars").glob("*.jar")),
    ]
)
```

jgoeders: This worked once I also included both the compiled files (/bin) and the distributed precompiled binaries (/jars)

jgoeders:
@dallinjdahl @reillymck OK, should be good to go now.

dallinjdahl (approved):
I like that better.  I don't know if the interchange project is pip-installable, but that might be a clean way going forward, to just have both on the requirements.txt.  I guess we could always download the interchange project separately and pip install rapidwright, but this works for now.

PR #114, opened 2023-02-28
Update bfasst to run WaFoVe correctly

JakeEdvenson:
@reillymck @dallinjdahl Would this be a possible implementation for allowing multiple arguments passed into flows.py? Or is it preferable to make multiple flows? I would expect at least 12 flows if this were the case to get most functionality working properly (although, the user would never be able to set both the seed and the number of tests to run since they can only set one comparison flag without the solution I'm proposing in this PR as far as I'm aware.)


dallinjdahl:
We discussed this a little on thursday.  Ideally we would have argparse configured in the base class, and options added for the individual tools, and then you just run argparse, instead of manually extracting substrings.  Does that make any sense?

JakeEdvenson:
@dallinjdahl Yeah I think that makes sense. So, when you say base class are you referring to run_design.py or base.py in the compare folder?

JakeEdvenson:
@dallinjdahl I found this on stack overflow, is this sort of what you're talking about? https://stackoverflow.com/questions/59086966/pythonic-way-for-adding-parameters-to-argparse-based-on-another-parameter

dallinjdahl:
The idea that we had was to set up options for the compare flow_args in the base.py, and then add tool-specific options in your specific tool class, and then call parse in your specific tool class, so that each tool has the general as well as the specific options configured.  We can discuss the sub-parser idea more, but we hadn't thought about that previously.

jgoeders:
@JakeEdvenson Is this ready to go now?  (I know some tests are failing, but that's due to changes I made)

JakeEdvenson:
@jgoeders Yup, the PR should be ready.

jgoeders (approved):

Issue #115, opened 2023-03-01
Reverse_bit_tool attribute error


JakeEdvenson:
I've tried running several flows using `python3 scripts/run_design.py examples/byu/alu xilinx_yosys_impl` (replacing the last part with other flow names) and whenever it reaches `status = xray_rev(design, build_dir, flow_args)` it throws an error at line 21 in scripts/bfasst/reverse_bit/base.py where it says `AttributeError: 'collections.defaultdict' object has no attribute 'read'`. I'm running this on the current commit from main in a virtual environment. Any thoughts?

reillymck:
This is an interesting bug. The flow function for xilinx_yosys_impl (I added the print statements):
```
def flow_xilinx_yosys_impl(design, flow_args, build_dir):
    """Vivado synthesis and implementation, reverse with xray, compare with yosys"""
    print(f"Full flow args: {flow_args}")
    status = vivado_synth(design, build_dir, flow_args)
    print(f"Full flow args1: {flow_args}")
    status = vivado_impl(design, build_dir, flow_args)

    # Run X-ray and fasm2bel
    print(f"Full flow args2: {flow_args}")
    status = xray_rev(design, build_dir, flow_args)

    status = yosys_cmp(design, build_dir, flow_args)
    return status
```
I also added print statements in the base classes for synth, impl, and reverse_bit to print the flow_args argument passed to the constructor. This is the print I get:
```
Full flow args: defaultdict(<class 'str'>, {<ToolType.SYNTH: 0>: '', <ToolType.IMPL: 1>: '', <ToolType.MAP: 2>: '', <ToolType.CMP: 3>: '', <ToolType.REVERSE: 5>: ''})
Synth Base flow args: 
Synthesis already run 
Full flow args1: defaultdict(<class 'str'>, {<ToolType.SYNTH: 0>: '', <ToolType.IMPL: 1>: '', <ToolType.MAP: 2>: '', <ToolType.CMP: 3>: '', <ToolType.REVERSE: 5>: ''})
Impl flow args: 
Implementation already run 
Full flow args2: defaultdict(<class 'str'>, {<ToolType.SYNTH: 0>: '', <ToolType.IMPL: 1>: '', <ToolType.MAP: 2>: '', <ToolType.CMP: 3>: '', <ToolType.REVERSE: 5>: ''})
[...]

PR #116, opened 2023-03-02
Add reverse to argparse

reillymck:
Parsing flow args was throwing an error because "reverse" was not part of the argparse namespace in run_design.py


reillymck:
Issue #115

jgoeders (approved):

PR #133, opened 2023-03-23
Created wafove test cases

JakeEdvenson:
@jgoeders Is this the type of experiments you're looking for with WaFoVe? I'm running them to confirm it passes all of the tests.


jgoeders:
@JakeEdvenson Looks like you need to add make install_wafove to https://github.com/byuccl/bfasst/blob/main/.github/actions/setup_env/action.yml

jgoeders:
@JakeEdvenson Looks like install wavofe asks for a sudo password.  Can you change it to add the required packages here https://github.com/byuccl/bfasst/blob/main/Makefile#L10 and not have your wafove install try and install packages?

jgoeders (approved):

PR #151, opened 2023-03-30
Wafove experiment fix

JakeEdvenson:
I believe that the error with the experiments process for wafove is due to pipelining the experiments. I found one of the commands being run in WaFoVe was not producing a file inside of the build directory, rather, it was being produced in the main cwd. As a result, the file could have been getting mixed up in a pipelined process and then leading to failures. At least, that's my current hypothesis.


jgoeders:
Ah that makes sense.  If files aren't getting put in the work directory they could clobber each other.  I'll merge this and we can see if it's stable.

jgoeders (approved):

PR #194, opened 2023-06-01
Removed redundant output from status string dunder

KeenanRileyFaulkner:


KeenanRileyFaulkner:
This is still messed up. My bad. Closing for now.

KeenanRileyFaulkner:
Reopened. Should work correctly now. Had to format the exception string differently, but it looks better.

PR #199, opened 2023-06-05
artifacts issue 187

KeenanRileyFaulkner:
- Added artifacts uploads for failing unit tests
- Fixed typo

### DIFF
--- a/.github/actions/setup_env/action.yml
+++ b/.github/actions/setup_env/action.yml
@@ -27,7 +27,7 @@ runs:
     run: |
       make install_fasm2bels
 
-  - name: fasm2bels_artifcats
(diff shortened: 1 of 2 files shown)


jgoeders:
@KeenanRileyFaulkner 
I don't see any artifact files here: https://github.com/byuccl/bfasst/actions/runs/5182421500?pr=199, which is fine since you have it marked for `if failure` (which I think is the correct way to go).  Have you tried removing the `if failure` to verify that the artifact files are generated correctly?

KeenanRileyFaulkner:
I think this is ready to merge now. Let me know if there are any additional changes to make.

Issue #152, opened 2023-03-30
run_design.py throws an error when the design is not within the designs folder


jgoeders:

jgoeders:
Copy a design to a directory outside of bfasst, or outside of the designs directory, and try running it with run_design.py.   It should error....

jgoeders:
@Ramez-git

dallinjdahl:
PR pending #200

PR #200, opened 2023-06-06
fixed Issue #152 run_design.py

Ramez-git:


third_party/RapidWright
jgoeders, line 1: Could you undo this change to the submodule version?

third_party/yosys
jgoeders, line 1: Could you undo this change to the submodule version?

scripts/run_design.py
```diff
@@ -23,7 +23,7 @@ def run_design(design_path, flow, error_flow, flow_args):
     design = Design(design_path)
 
     # Create temp folder
-    build_dir = Path.cwd() / "build" / flow / (design.path.relative_to(paths.DESIGNS_PATH))
+    build_dir = Path.cwd() / "build" / flow / design.path.name
```
jgoeders, line 26: I think this will work to support designs outside of the examples directory, however it removes the current directory organization within the build directory.

Perhaps a better solution would be to check if it with the designs directory (https://stackoverflow.com/a/34236245/609215) and in that case leave the existing directory structure.  

If it outside of the designs directory, perhaps we could do something like:
`"build" / flow / "<external>" / design.path.name`

@dallinjdahl Thoughts?

jgoeders (changes requested):

scripts/run_design.py
```diff
@@ -23,7 +23,11 @@ def run_design(design_path, flow, error_flow, flow_args):
     design = Design(design_path)
 
     # Create temp folder
-    build_dir = Path.cwd() / "build" / flow / (design.path.relative_to(paths.DESIGNS_PATH))
+    print(Path.cwd())
+    print(flow)
+    print(design.path)
+    print(paths.DESIGNS_PATH)
+    build_dir = Path.cwd() / "build" / flow / design.path.parent #change made here
```
jgoeders, line 30: I don't think this is the right way to fix it.  I think we want to keep the existing `relative_to` code.  But in the case that the design_path is not a subdirectory of DESIGNS_PATH, then we should handle it differently, and we could use `Path.cwd() / "build" / flow / "<external>" / design_path.name`

jgoeders (changes requested):

Makefile
```diff
@@ -70,7 +70,7 @@ install_fasm2bels: submodules
 	$(IN_ENV) cd third_party/fasm2bels && make test-py
 
 # Run a simple design through fasm2bels to generate the database.
-	$(IN_ENV) python scripts/run_design.py designs/basic/and3/ xilinx_and_reversed
+	$(IN_ENV) python scripts/run_design.py basic/and3/ xilinx_and_reversed
```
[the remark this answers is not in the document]
Ramez-git: This was causing an error when running make install log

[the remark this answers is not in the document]
Ramez-git: in paths.py line 7 it already includes designs in the path thus in design.py the relative path won't compute

scripts/run_design.py
dallinjdahl: I think if we're going to move to support external designs with arbitrary file paths we should move the build artifacts next to each design. If we use the external tag, we risk overwriting designs with the same name but different paths.

jgoeders: I agree <external> is a bit of a hack, but I also rather not put the generated files beside the source code. They did this in another project I'm a part of and I'm not a fan. It got really messy. 

In this case, I rather reconstruct the full absolute path within the build folder.

dallinjdahl: given an arbitrary path `/path/to/source` we could derive `/path/to/build/artifact` such that all the artifacts are still in a separate build directory, but relative to the source rather than our tool.  That matches the way almost all tools work.  I don't know that I can think of a tool that depends on paths relative to the install directory...

Makefile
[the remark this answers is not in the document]
jgoeders: I can't reproduce this issue on my machine.  I just ran:
```
python scripts/run_design.py designs/basic/and3/ xilinx_and_reversed
```
from main and it worked fine.  

`design.path` is an absolute path (https://github.com/byuccl/bfasst/blob/dcaa1e32324f4f7b1527e0d72ce1e69ab2b67033/bfasst/design.py#L28) and calling  `.relative_to(paths.DESIGNS_PATH)` on it should be fine:

```
Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pathlib
>>> from bfasst import paths
>>> p = pathlib.Path("designs/basic/and3/").absolute()
>>> print(p)
/home/jgoeders/bfasst/designs/basic/and3
>>> p.relative_to(paths.DESIGNS_PATH)
PosixPath('basic/and3')
```

scripts/run_design.py
jgoeders: The build directory is not necessarily placed in the install directory.  It is placed in cwd (https://github.com/byuccl/bfasst/blob/d3bc3b1f3969326d38c72297072a3880ba3f9795/scripts/run_design.py#L26 -- although this script probably ought to have a command line option to override the location).

Makefile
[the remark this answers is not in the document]
Ramez-git: then that must mean the issue is from my machine

PR #203, opened 2023-06-07
Created weekly CI check to test large designs

KeenanRileyFaulkner:


experiments/weekly_tests/xilinx.yaml
```diff
@@ -0,0 +1,5 @@
+design_dirs:
+  - ooc/
```
reillymck, line 2: I bunch of the ooc designs will fail the xilinx flow because their ports cannot automatically be mapped to I/O pins. I think for the xilinx flow we should just do base/ designs and for xilinx_ooc we run just the ooc designs.

reillymck (changes requested):

jgoeders:
@KeenanRileyFaulkner I've never really set up a weekly CI like this....  What happens when it fails.  Do you know if it will email me?  Normally the CI is attached to a PR and I think it emails the PR author when it fails.  Is there a way to have this email me if it fails?

Also, can you add a "badge" to the readme with the weekly runs status?

Issue #207, opened 2023-06-09
Use new rapidwright wrapper script to download device files


jgoeders:
Rapidwright now has a convenient wrapper script that can be used to invoke it directly without having to worry about Java classpath.  We can simply how we download the device files and remove the extra python script we have:

https://github.com/Xilinx/RapidWright/issues/669#event-9476001168  

[This ](https://github.com/byuccl/bfasst/blob/37c7ddd2e71883ce9eef6560adf56421749752b6/Makefile#L65) call and it's associate script could be replaced by the command in the issue linked above.

PR #213, opened 2023-06-13
#207 new rapidwright wrapper

Tscook17:
I changed the rapidwright makefile target to use the new rapidwright wrapper.


Makefile
```diff
@@ -62,7 +62,8 @@ submodules:
 rapidwright: submodules
 	cd third_party/RapidWright && ./gradlew compileJava
 	cd third_party/RapidWright/interchange/ && make
-	$(IN_ENV) python scripts/rapidwright_get_data_files.py
+	$(IN_ENV) cd third_party/RapidWright && ./gradlew compileJava ; export PATH=`pwd`/bin:$$PATH ; \
```
jgoeders, line 65: I don't think we need the `&& ./gradlew compileJava` part as we have it already up on line 63.

jgoeders (changes requested):

PR #216, opened 2023-06-14
Structural compare: check nets

jgoeders:


jgoeders:
Breaking because of #217

PR #218, opened 2023-06-15
status issue 127

KeenanRileyFaulkner:
- Added for help debugging
- Fixed status problem by resolving circular dependency

### DIFF
--- a/bfasst/status.py
+++ b/bfasst/status.py
@@ -109,6 +109,7 @@ class BfasstException(Exception):
 
 class Status:
     """Represents the status of a BFASST tool after it has run"""
+
     def __init__(self, status, msg="", raise_excep=True):
         self.status = status
         self.msg = msg if msg else ""
--- a/bfasst/status.py
+++ b/bfasst/status.py
@@ -97,6 +97,9 @@ msg_map = {
 
 
 class BfasstException(Exception):
+    # The creator is the object that raised the exception
+    # If the creator is not none, then the exception was raised from a bfasst
+    # status object
     def __init__(self, err, msg, creator=None):
         super().__init__(msg)
         self.error = err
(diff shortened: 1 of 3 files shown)


bfasst/status.py
```diff
@@ -97,18 +97,25 @@ class MapStatus(enum.Enum):
 
 
 class BfasstException(Exception):
-    def __init__(self, err, msg):
+    def __init__(self, err, msg, creator=None):
```
jgoeders, line 100: Can you add a comment here explaining this argument?

bfasst/status.py
```diff
@@ -97,18 +97,25 @@ class MapStatus(enum.Enum):
...
-            raise BfasstException(status, f": {self.msg}")
+            if msg and msg not in msg_map[self.status]:
+                raise BfasstException(status, f"{msg_map[self.status]}: {self.msg}", self)
+            raise BfasstException(status, msg_map[self.status], self)
 
     def __str__(self):
-        return f"{msg_map[self.status]}{self.msg}"
+        if self.msg not in msg_map[self.status]:
+            return f"{msg_map[self.status]}: {self.msg}"
+        return msg_map[self.status]
```
jgoeders, line 121: Can you add some comments here explaining why this approach was necessary?  Thanks!

jgoeders (changes requested):

bfasst/status.py
KeenanRileyFaulkner: Done

bfasst/status.py
KeenanRileyFaulkner: Done

scripts/run_experiment.py
```diff
@@ -72,7 +72,11 @@ def run_design(print_lock, running_list, design, work_path, flow_fcn, flow_args)
     try:
         status = flow_fcn(design=design, build_dir=work_path, flow_args=flow_args)
     except BfasstException as e:
-        status = Status(status=e.error, msg=str(e), raise_excep=False)
+        status = (
```
KeenanRileyFaulkner, line 75: The reason this is necessary, instead of just returning part of the exception (eg e.error) is because run_experiment needs to catch the exception and continue to run other designs. run_design, on the other hand, just lets the exception go and it prints to the terminal. This is also the reason behind the two separate if blocks on lines 120 and 127 of bfasst/status.py @jgoeders

Issue #224, opened 2023-06-16
Remove bfasst/error_injection and all references to it


KeenanRileyFaulkner:
Remove the old bfasst error injector from the project, now that we have integrated a new one in that uses spydrnet. The old injector is found at [`bfasst/error_injection`.](https://github.com/byuccl/bfasst/tree/e3e4b4f650a3dd364b5c30a14f2c0eb5bdbb14e1/bfasst/error_injection) **DO NOT REMOVE THE INJECTOR FOUND AT `bfasst/transform`**
- [x] Remove the tool
- [x] Remove the tool_wrapper methods that instance the tool
- [x] Remove the flows that call the tool_wrapper methods
- [x] Clean up associated imports
- [x] Remove the alias from the import for the _new error injector_ in [`flows.py`](https://github.com/byuccl/bfasst/blob/e3e4b4f650a3dd364b5c30a14f2c0eb5bdbb14e1/bfasst/flows.py#L34) and [unalias the instance of the new error injector](https://github.com/byuccl/bfasst/blob/e3e4b4f650a3dd364b5c30a14f2c0eb5bdbb14e1/bfasst/flows.py#L252)

KeenanRileyFaulkner:
This has been done within flow_refactor and will be part of #228

Issue #223, opened 2023-06-16
Allow for flows to return subflows that need to be ran


reillymck:
- [x] Add job class with function pointer to its job and a list of the jobs it depends on
- [x] Change flows to create job objects to be executed.
- [x] Run experiment will run Flow.create() as preprocessing to build a list of jobs that needs done
- [x] Run experiment will cycle through the job list, running jobs with no dependencies as threads become available. When a thread finishes, run experiment removes that job from the remaining jobs' dependency lists (or set).
- [x] Run design will run jobs in the list consecutively (no need to worry about dependencies if job list is guaranteed to be in at least hierarchal order).
- [x] Run experiment/run design checks for BfasstException and prints errors/success messages accordingly

Issue #221, opened 2023-06-16
Move Flows To Classes


KeenanRileyFaulkner:
In order to allow for easier parallel processing in run experiment, it would be good if flows were classes instead.

- [x] Create Flow abstract class with abstract run() function.
- [x] Moves flows to their own class (also absorb tool wrappers into flows (although xilinx_ooc will be a subflow)).

- [x] Create exceptions for each tool that inherit from BfasstException. Stop returning statuses for tools and instead raise exceptions
- [x] Edit Status class so that it only relates to flows (no exception raising). Flows will return a status. The Flow.run() will have one try catch clause for all the tools that catches the parent BfassException, and passes the exception type to a failed status.

PR #228, opened 2023-06-19
Flow refactor

reillymck:


reillymck:
Needs exception refactor to work properly, among other small adjustments.

bfasst/flows/xilinx_structural_error_injection_flow.py
```diff
@@ -0,0 +1,70 @@
...
+
+                # Reset the mappings in the compare tool, give it a number for logging the run,
+                # and run the comparison
+                compare_tool.reset_mappings()
+
+                # The compare tool should only reset its logs once per flow,
+                # not per error type, but should be unaware of what type of error is being injected
+                compare_tool.run_num = i if err == injection_types[0] else i + num_runs
+
+                if compare_tool.compare_netlists() is None:
```
KeenanRileyFaulkner, line 57: This may not be what we need, since structural.compare will raise an exception if the two netlists are not equivalent. If _no exception is raised_ by `compare_netlists()`, we've got a problem and should run the code in this if block.

reillymck:
Fixes #223

bfasst/flows/flow.py
```diff
@@ -0,0 +1,62 @@
...
+    def create(self):
+        """Create all the jobs necessary for the flow"""
+        raise NotImplementedError
+
+    def _get_first_job(self):
+        if len(self.job_list) > 0:
+            return self.job_list[0]
+        return None
+
+    def modify_first_job_dependencies(self, new_dependencies: list):
```
jgoeders, line 53: Can you add a docstring to this function?  I'm not sure from the name what it's purpose is, and it seems like an important function.

bfasst/flows/gather_impl_data_flow.py
```diff
@@ -0,0 +1,135 @@
+"""GatherImplData flow"""
```
jgoeders, line 1: Can you expand this docstring?  I'm also a bit confused on the naming of this.  What does "gather implementation" mean?

bfasst/flows/gather_impl_data_flow.py
jgoeders, line 1: If these are in the `flows` folder, I think we can remove the *_flow* from the names.

bfasst/flows/sub_flows/conformal_only_flow.py
jgoeders, line 1: Can we just call this conformal.py instead of conformal_only_flow.py?

bfasst/job.py
```diff
@@ -0,0 +1,34 @@
...
+from bfasst.tool import BfasstException
+
+
+class Job:
+    """Class for a job function and dependency list"""
+
+    def __init__(self, function, design_rel_path, dependencies=None):
+        self.function = function
+        self.design_rel_path = design_rel_path
+        self.dependencies = dependencies
```
jgoeders, line 14: Can you comment these?

bfasst/job.py
```diff
@@ -0,0 +1,34 @@
...
+        self.__set_uuid()
+
+    def __set_uuid(self):
```
jgoeders, line 17: Can you expand this docstring?  What's the purpose of the UUID?  Why not just use the job instance itself as an identifier?

bfasst/opt/base.py
```diff
@@ -1,22 +1,26 @@
...
 
 
 class OptTool(Tool):
     """Base class for logic optimization tools"""
 
-    success_status = Status(OptStatus.SUCCESS)
-
     def __init__(self, cwd, design, flow_args="") -> None:
         super().__init__(cwd, design)
+        self.yosys_netlist_path = [str(self.design.netlist_path)]
```
jgoeders, line 18: Why does opt base have a property named *yosys*? That seems incorrect...

jgoeders (changes requested):

bfasst/job.py
KeenanRileyFaulkner: We can't use the job instance because the jobs were having problems pickling between processes. A uuid allows us to look up the job in the job list when we need to run a callback (like cleaning up the job list after a job finishes) without having to pickle the job. Hunting down all the pickling errors was a nightmare

jgoeders: Sounds good.  Just add a bit of that explanation to the docstring.

bfasst/flows/flow.py
KeenanRileyFaulkner: Done

bfasst/job.py
KeenanRileyFaulkner: Done

bfasst/job.py
KeenanRileyFaulkner: Done

bfasst/opt/base.py
KeenanRileyFaulkner: It is used here in the implementing classes. I'm not the one who wrote this, so if naming is wrong, I don't know why: https://github.com/byuccl/bfasst/blob/e865755c6243aef6ef3b31b6ca143587bcf78bad/bfasst/opt/ic2_lse.py#L47
https://github.com/byuccl/bfasst/blob/e865755c6243aef6ef3b31b6ca143587bcf78bad/bfasst/opt/ic2_synplify.py#L51

jgoeders: Ok, don't worry about it then.

bfasst/flows/sub_flows/conformal_only_flow.py
KeenanRileyFaulkner: Done

bfasst/flows/gather_impl_data_flow.py
KeenanRileyFaulkner: Done

KeenanRileyFaulkner:
Resolves #221, resolves #223, resolves #224

PR #243, opened 2023-07-11
Update run_design.py

Ramez-git:
#200

### DIFF
--- a/scripts/run_design.py
+++ b/scripts/run_design.py
@@ -6,47 +6,41 @@ from pathlib import Path
 
 from bfasst.design import Design
 from bfasst import paths
-from bfasst.flows.flow import get_flow, get_flows
-from bfasst.tool import BfasstException
-from bfasst.types import ToolType
+from bfasst.flows import Flows, run_flow, get_flow_fcn_by_name
+from bfasst.tool_wrappers import ToolType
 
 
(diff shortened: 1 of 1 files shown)


scripts/run_design.py
```diff
@@ -6,47 +6,41 @@
 
 from bfasst.design import Design
 from bfasst import paths
-from bfasst.flows.flow import get_flow, get_flows
```
KeenanRileyFaulkner, line 9: We have refactored the whole code base to allow flows to be split up and run on separate threads if necessary. It looks like you may have forgotten to sync your fork before you made changes, because most of these changes revert back to the old way of doing things. The design object at this point is self contained, so this `run_design.py` file no longer needs to be modified at all. You should explore `bfasst/design.py` [here](https://github.com/byuccl/bfasst/blob/main/bfasst/design.py) to determine where to make the required changes. 

I think you can change [line 34 in `bfasst/design.py`](https://github.com/byuccl/bfasst/blob/203c5c782d97e8b0b026f6f3500574b48860b549/bfasst/design.py#L34) to `self.rel_path = self.path.relative_to(paths.DESIGNS_PATH) if paths.DESIGNS_PATH in self.path.parents else self.path` in order to effectively make the build and design dir the same (which may or may not be the desired behavior). This will also allow the run_design script to work for any design at any location.

Issue #238, opened 2023-07-05
Issue with physical netlist generation for byu/riscv_final


KeenanRileyFaulkner:
Line 786 in bfasst is breaking the weekly error injection CI. However, the same unit tests pass. The issue is most likely to be with designs that are part of the error_injection weekly test but not part of the physical_netlist unit test, since the error_injection test more comprehensively tests the designs in the byu design directory.
https://github.com/byuccl/bfasst/suites/14002997756/artifacts/781358036

KeenanRileyFaulkner:
Successfully reproduced. The error only occurs in the riscv_final design during the physical netlist generation:
```
Running: riscv_fi (0:02:00) concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.10/concurrent/futures/process.py", line 246, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/keenanrf/research/test/bfasst/scripts/run_experiment.py", line 148, in run_job
    job.function()
  File "/home/keenanrf/research/test/bfasst/bfasst/transform/xilinx_phys_netlist.py", line 90, in run
    self.run_rapidwright(phys_netlist_checkpoint, phys_netlist_edif_path)
  File "/home/keenanrf/research/test/bfasst/bfasst/transform/xilinx_phys_netlist.py", line 160, in run_rapidwright
    self.process_all_luts(cells_already_visited)
  File "/home/keenanrf/research/test/bfasst/bfasst/transform/xilinx_phys_netlist.py", line 269, in process_all_luts
    self.process_lut(lut6_cell, lut5_cell)
  File "/home/keenanrf/research/test/bfasst/bfasst/transform/xilinx_phys_netlist.py", line 645, in process_lut
    self.lut_move_net_to_new_cell(
  File "/home/keenanrf/research/test/bfasst/bfasst/transform/xilinx_phys_netlist.py", line 786, in lut_move_net_to_new_cell
    new_logical_pin = f"I{int(str(physical_pin[1])) - 1}"
ValueError: invalid literal for int() with base 10: 'L'
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/keenanrf/research/test/bfasst/scripts/run_experiment.py", line 263, in <module>
    main(args.experiment_yaml, args.threads, args.print_period)
  File "/home/keenanrf/research/test/bfasst/scripts/run_experiment.py", line 83, in main
[...]

Issue #242, opened 2023-07-11
Run Experiment is not actually killed by ctrl+c or by killing master terminal.


reillymck:

KeenanRileyFaulkner:
@reillymck now that we have ninja merged, should this issue be closed?

PR #245, opened 2023-07-14
LUTRAM fixes

reillymck:
Addresses the problem outlined in #238 concerning LUTRAMs. Also adds more LUTRAM support, support for LUT5 + gnd lut being on the same LUT6_2.


reillymck:
Drafting to see unit test results

reillymck:
Soft fix for #242 -> When Ctrl-C is intercepted, the pool is given a immediate kill signal. From what I understand, it won't kill futures that have already started, but these should finish up shortly after the keyboard interrupt. There does not seem to be a straightforward way from the multiprocessing library to send os signals to running futures.

bfasst/ninja_flows/vivado.py
```diff
@@ -0,0 +1,32 @@
...
+    def create_build_snippets(self):
+        self.vivado_tool.create_build_snippets()
+
+    def add_ninja_deps(self, deps=None):
+        if not deps:
+            deps = []
+        deps.extend(self.vivado_tool.add_ninja_deps())
+        deps.append(f"{NINJA_FLOWS_PATH}/vivado.py ")
+        if self.ooc:
+            deps.append(f"{NINJA_FLOWS_PATH}/vivado_ooc.py ")
```
jgoeders, line 26: Are deps always paths?  If so, we should use `Path` not `str`

bfasst/utils/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Import general functions into the utils namespace"""
+from .general import *
```
jgoeders, line 2: I'm not sure about this approach.  Isn't it better to just import utility functions as needed?

Otherwise you have code that does this:
```python
import bfasst.utils

<use function where it's not clear where it originated>
```

designs/ooc/data_path/LIFO.v
```diff
@@ -45,10 +45,10 @@ module LIFO(
...
 			 end
 		
-	always @(posedge clk)
-      if (wr_en)
-         ram[addr] <= din;
-
+	always @(posedge clk)
+      if (wr_en)
+         ram[addr] <= din;
+
```
jgoeders, line 51: Do we want to change this permanently?

designs/ooc/mpeg2fpga/framestore.v
```diff
@@ -398,7 +398,7 @@ module framestore(rst, clk, mem_clk,
     #(.addr_width(MEMRESP_DEPTH),
     .dta_width(9'd64),
     .prog_thresh(MEMRESP_THRESHOLD),   // threshold to make mem_ctl stop writing before mem_response_fifo overflows
-    .FIFO_XILINX(1))
+    .FIFO_XILINX(0))
```
jgoeders, line 401: Do we want to change this permanently?

designs/ooc/natalius_8bit_risc/LIFO.v
```diff
@@ -45,10 +46,10 @@ module LIFO(
...
 			 end
 		
-	always @(posedge clk)
-      if (wr_en)
-         ram[addr] <= din;
-
+	always @(posedge clk)
+      if (wr_en)
+         ram[addr] <= din;
+
```
jgoeders, line 52: Do we want to change this permanently?

designs/vtr_benchmarks/synth_design.tcl
jgoeders, line 1: What's this file used for?  ...I haven't seen a tcl script located with benchmarks before....

jgoeders (changes requested):

bfasst/ninja_flows/ninja_flow_manager.py
```diff
@@ -0,0 +1,87 @@
...
+                    "deps": self.flows[0].add_ninja_deps(),
+                    "flow": self.flow_name,
+                    "flow_args": self.flow_args,
+                },
+            )
+
+        return master_ninja
+
+    def __get_absolute_design_paths(self):
+        self.designs = [str(DESIGNS_PATH / design) for design in self.designs]
```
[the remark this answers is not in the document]
KeenanRileyFaulkner: Yes, we are templating these paths into the master.ninja.mustache file. All templated types must be basic primitives (string, int, etc.) so path types won't work.

bfasst/ninja_flows/vivado.py
KeenanRileyFaulkner: They are paths, but they are templated into the build.ninja file, and so need to be strings, not paths. These deps are used to determine when to rebuild the build.ninja file. If any of them change, the build.ninja file will rerender itself. But we have to template them into the file as strings.

bfasst/ninja_flows/ninja_flow_manager.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def get_design_basenames(designs):
+    return [("/").join(design.split("/")[-2:]) for design in designs]
```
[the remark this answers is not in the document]
KeenanRileyFaulkner: In this case, because we need the path as a string anyways, doesn't it make more sense to leave this syntax as is, rather than try to convert to a path and back with `str(Path(*path.parts[-2:]))`?

bfasst/ninja_flows/ninja_flow_manager.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class NinjaFlowManager:
+    """Utility to manage the creation and execution of ninja flows."""
+
+    def __init__(self):
+        self.flows = None
+        self.designs = None
+        self.flow_name = None
+        self.flow_args = None
```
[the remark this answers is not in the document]
KeenanRileyFaulkner: Done

PR #265, opened 2023-08-01
resolved issue #152

Ramez-git:


bfasst/design.py
```diff
@@ -23,15 +23,22 @@ class Design:
     "class holding paths and other metadata for a given design" ""
 
     def __init__(self, dir_path, experiment_dir):
-        if not dir_path.is_dir() and (paths.DESIGNS_PATH / dir_path).is_dir():
+        if not dir_path.is_dir() and dir_path.is_absolute():
+            dir_path = dir_path
```
jgoeders, line 27: This part of the "if" logic isn't doing anything.  What is it meant to accomplish?

bfasst/design.py
```diff
@@ -23,15 +23,22 @@ class Design:
...
+        elif not dir_path.is_dir() and (paths.DESIGNS_PATH / dir_path).is_dir():
```
jgoeders, line 28: Perhaps you want to add `not dir_path.is_absolute()` to this if statement?

bfasst/design.py
```diff
@@ -23,15 +23,22 @@ class Design:
...
             dir_path = paths.DESIGNS_PATH / dir_path
-        dir_path = dir_path.absolute()
+
+        dir_path = paths.ROOT_PATH / dir_path.absolute()
```
jgoeders, line 31: This doesn't look correct.  The right side of a path "/" operation should not be an absolute path.  What are you trying to accomplish with this line?

bfasst/design.py
```diff
@@ -23,15 +23,22 @@ class Design:
...
 
         if not dir_path.is_dir():
             error("Design folder", dir_path, " does not exist.")
 
         self.path = dir_path
-        self.rel_path = self.path.relative_to(paths.DESIGNS_PATH)
+        if self.path.relative_to(paths.DESIGNS_PATH) != None:
```
jgoeders, line 37: Don't use `!= None` in Python.  You can use `is not None` or in this case, it's probably fine to just say `if self.path.relative_to(paths.DESIGNS_PATH):`

scripts/run_design.py
```diff
@@ -90,8 +92,15 @@ def main():
     for arg_name, enum in flow_args_map.items():
         flow_args[enum] = getattr(args, arg_name)
 
-    run_design(Path(args.design_path), args.flow, args.error_flow, flow_args=flow_args)
+    # Convert the design path to an absolute path
+    design_path = Path(os.path.expanduser(args.design_path))
```
jgoeders, line 96: What is the call to `expanduser` meant to do?  This function expands the `~` in a path, but it would have already been done by the shell prior to this code running, so I'm not sure what it's accomplishing here?

scripts/run_design.py
```diff
@@ -63,6 +64,7 @@ def main():
     parser.add_argument("--reverse", help="Reverse args", type=str, default="")
     parser.add_argument("--err", help="Error flow args", type=str, default="")
     parser.add_argument("--quiet", action="store_true")
+    parser.add_argument("--experiment_dir", help="Experiment directory", type=str, default=None)
```
jgoeders, line 67: The `type` should be `pathlib.Path` if the user is providing a path.

scripts/run_design.py
```diff
@@ -90,8 +92,15 @@ def main():
     for arg_name, enum in flow_args_map.items():
         flow_args[enum] = getattr(args, arg_name)
 
-    run_design(Path(args.design_path), args.flow, args.error_flow, flow_args=flow_args)
+    # Convert the design path to an absolute path
+    design_path = Path(os.path.expanduser(args.design_path))
+
+    # Set experiment_dir to the current working directory if not provided
+    experiment_dir = Path.cwd() if args.experiment_dir is None else Path(os.path.expanduser(args.experiment_dir))
```
jgoeders, line 99: Again, I don't think you need to `expanduser` as that should have already been done by the shell?

jgoeders (changes requested):

bfasst/design.py
Ramez-git: reverted

bfasst/design.py
Ramez-git: reverted

bfasst/design.py
Ramez-git: reverted

scripts/run_design.py
Ramez-git: reverted

scripts/run_design.py
Ramez-git: reverted

scripts/run_design.py
Ramez-git: reverted

bfasst/design.py
Ramez-git: reverted

jgoeders (approved):

PR #274, opened 2023-08-23
updated to build in cwd where script is invoked

KeenanRileyFaulkner:


KeenanRileyFaulkner:
The CI checks are failing due to env setup problems, but this is ready to merge, and has been tested locally. It will build in the cwd when the bfasster script is invoked. If the same experiment is invoked in the same directory (aka the artifacts exist in their proper ./build/<design>/... location) and the artifacts are up to date, ninja will not rebuild them, otherwise it will.

jgoeders:
OK, I'm working on merging a fix for the environment.

jgoeders (approved):

bfasst/yaml_parser.py
```diff
@@ -63,7 +60,7 @@ def __collect_design_paths(self):
                         continue
 
                     if (design_child / "design.yaml").is_file():
-                        design_name = design_child.split("/")[-2:]
+                        design_name = "/".join(str(design_child).split("/")[-2:])
```
jgoeders, line 63: This seems to be removing the first to items of the path?  What's the reason for this?  

It seems like you might want to use `relpath`, relative to something?

Or if you really want to always remove the first to elements, add a comment explaining why, and there is a better approach here: https://stackoverflow.com/a/26724369/609215

jgoeders (changes requested):

bfasst/yaml_parser.py
KeenanRileyFaulkner: Basically, this is because all the ninja stuff right now expects to take the design in as a subdirectory of `designs/`. This reverts the change Ramez made a few weeks ago and is undesirable, but it is in line with the current structure of our CI yaml experiments. Reilly and I have already talked about this, and we want to make the parser require the full relative path to each design we want to run. This is planned for a future pr. I was trying to get to it this week, but have been really bogged down by rapidwright stuff.

KeenanRileyFaulkner: The CI checks right now specify something like:
```
designs:
    - byu/alu
```
And we want to make it require
```
designs:
    - designs/byu/alu
```
Or some other complete path relative to where the script is invoked. This will allow designs to reside outside the bfasst repo. The bfasster script allows multiple designs to be passed by yaml or a single design to be passed by cmd line argument, so the format for the designs in both cases needs to match. I initially made the format match what was already in the CI yaml files.

KeenanRileyFaulkner: I just made the requested changes, in the interim.

PR #343, opened 2023-11-13
Moved configure_vivado_synth_tools to flow_utils.py

### LINKED ISSUE #314: Move `configure_vivado_tool` out of `Flow` class

yonnorc42:
Fixes #314


bfasst/ninja_flows/flow_utils.py
```diff
@@ -30,3 +32,9 @@ def create_build_file():
     """Overwrite the build.ninja file when running generation and build steps separately"""
     with open(NINJA_BUILD_PATH, "w") as f:
         f.write("# This file is autogenerated by bfasst\n\n")
+
+
+def configure_vivado_synth_tool(design, flow_args, ooc=False):
+    if isinstance(flow_args, dict) and "synth" in flow_args:
+        return VivadoSynth(design, flow_args["synth"], ooc)
+    return VivadoSynth(design, ooc=ooc)
```
jgoeders, line 40: It seems this function is just checking if `"synth"` is in `flow_args`?

We could entirely bypass the need for this function by making sure `flow_args` was a [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict)

Then, anywhere we are calling this function, we could simply call 

```python
VivadoSynth(design, flow_args["synth"], ooc)
```

And it would work regardless of whether "synth" was a key in the dict or not.  

I think this would be more readable than calling into a separate function to create the object.

jgoeders (changes requested):

bfasst/ninja_flows/flow_utils.py
yonnorc42: To do this, would I go to the place where flow_args is initialized and make it a defaultdict instead of a dict? And then from there I could replace everywhere configure_vivado_synth_tool(design, flow_args, ooc) is called with VivadoSynth(design, flow_args["synth"], ooc)?

jgoeders: @yonnorc42 

Yes, exactly.

reillymck: Also, you could use flow_args.get("synth"). According to the [docs](https://docs.python.org/3.10/library/stdtypes.html?highlight=dict#mapping-types-dict): the `dict.get(key)` will return None if the key does not exist, which is the default value for the flow_args parameter in the VivadoSynth constructor. I think this may be better instead of a default dict? Also @KeenanRileyFaulkner, I see there is an `isinstance()` check on flow args - is there ever a case where it is not a dict? 
I think we could just replace every call to `configure_vivado_synth_tool` with `VivadoSynth(design, flow_args.get("synth"), ooc=ooc)`

reillymck: My understanding is the defaultdict woult prevent a KeyError from ever being thrown on a bad access. I'm not sure if we want to totally remove that functionality, in my opinion the use of dict.get() ensures that in this one place we are okay with it returning None instead of throwing a KeyError on a missing key.

jgoeders: I agree.

reillymck:
This is looking pretty good! If you would just delete the `configure_vivado_tool` function from flow_utils then it should be good to go. (pending the unit tests passing).

KeenanRileyFaulkner:
I think any issues with the dictionary instance, if they exist, will come up because of how our configure rule works. It has to blow away the ninja build file and remake it. Potential issues would be if we change a script file and run ninja again, and the issue would have to do with how the flow args are re-read into the build file as seen at the bottom of the ninja_flow_manager.py file

yonnorc42:
It looks like one of the unit tests (so far) failed because of the .get("synth")

jgoeders:
@yonnorc42 OK, looks like we will need a few more fixes to resolve this.  In places like [this](https://github.com/byuccl/bfasst/blob/d6a76954d2f9b9eddf2197ec4ce87b547ebf9b54/bfasst/ninja_flows/vivado_yosys_impl.py#L11), you will need to do something like:

```python
if flow_args is None:
  flow_args = []
```

You may be tempted to change the default `flow_args` argument value to `[]`, but [this is a bad idea](https://nikos7am.com/posts/mutable-default-arguments/).

yonnorc42:
> @yonnorc42 OK, looks like we will need a few more fixes to resolve this. In places like [this](https://github.com/byuccl/bfasst/blob/d6a76954d2f9b9eddf2197ec4ce87b547ebf9b54/bfasst/ninja_flows/vivado_yosys_impl.py#L11), you will need to do something like:
> 
> ```python
> if flow_args is None:
>   flow_args = []
> ```
> 
> You may be tempted to change the default `flow_args` argument value to `[]`, but [this is a bad idea](https://nikos7am.com/posts/mutable-default-arguments/).

In this case, wouldn't I make it an empty dictionary instead of an empty list?

jgoeders:
@yonnorc42 

> In this case, wouldn't I make it an empty dictionary instead of an empty list?

Yes, you're correct.

yonnorc42:
> @yonnorc42 OK, looks like we will need a few more fixes to resolve this. In places like [this](https://github.com/byuccl/bfasst/blob/d6a76954d2f9b9eddf2197ec4ce87b547ebf9b54/bfasst/ninja_flows/vivado_yosys_impl.py#L11), you will need to do something like:
> 
> ```python
> if flow_args is None:
>   flow_args = []
> ```
> 
> You may be tempted to change the default `flow_args` argument value to `[]`, but [this is a bad idea](https://nikos7am.com/posts/mutable-default-arguments/).

When I implement that if statement into the required files, it passes the unittests now, but it fails the pylint tests, and I think it's saying I failed because I have duplicate lines of code in all those places. This makes me think it wants me to use a function instead of a duplicate if statement in all those files, but I feel like that's essentially what we had originally with configure_vivado_synth_tool

yonnorc42:
I think this is ready to merge, unless you want me to change it

jgoeders (approved):

PR #368, opened 2023-11-28
Make sure structural compare fails on error and fix current comparison bug

### LINKED ISSUE #367: Structural Comparison Failing
Structural Comparisons in the `vivado_phys_netlist_cmp` flow are failing with the following error: `FAIL: Unhandled properties for type GND`.

Note, due to the way structural comparisons are used in the `vivado_structural_error_injection` flow, the compare tool will not report failures to ninja. So running a command line like `python scripts/run.py vivado_phys_netlist_cmp byu/alu` will appear to succeed, but checking the associated `struct_cmp.log` file will show the above error.

### LINKED ISSUE #370: Logs not cleared on subsequent runs of vivado_phys_netlist_cmp, structural comparison

### LINKED ISSUE #371: Netlist cleanup does not remove unused IBUFs

jgoeders:
Fixes #367 
Fixes #370 
Fixes #371


bfasst/ninja_utils/structural.py
```diff
@@ -771,5 +775,4 @@ def get_netlist(self, library):
         logging.info("SUCCESS")
     except StructuralCompareError as e:
         logging.error("FAIL: %s", e)
-
-    # pylint: enable=duplicate-code
+        sys.exit(1)
```
KeenanRileyFaulkner, line 778: One issue with this is that it will probably cause the error injection flow to fail, since the corrupt netlists won't match the original one and will now trigger ninja to raise an exception and stop running. There were some issues with simply catching those exceptions in the error injector flow and inverting them (raising if nothing was raised) that caused us to abandon that approach, but I can't remember what those issues were right now. @dallinjdahl might have more insight...?

jgoeders: OK, thanks.  I'll take a look at this.

dallinjdahl: The problem with this was that we wanted to be able to use the structural comparison tool as a generic comparator.  The fact that the comparison yields that they are not identical is not an error, nor is the fact that they were identical.  It's up to a later stage of the process to decide what the results should be, and error out if necessary.

We settled on having nonzero exit status indicate that something went wrong while running the script.  That way, we can run ninja with full parallelism, and it will stop on the first error, as it's something that needs to be debugged.  Otherwise we had issues stopping ninja after a certain number of failures, and having to kill several batches of queued jobs.

jgoeders: @KeenanRileyFaulkner The fix I've pushed updates the structural Tool to accept an argument specifying whether the comparison is expected to match or not.  This allows the flow to control the failure behavior.

jgoeders:
@KeenanRileyFaulkner Would you be able to look at these error injection failures? https://github.com/byuccl/bfasst/actions/runs/7021895761/job/19106142210?pr=368

PR #372, opened 2023-11-28
Removing flow_args

jgoeders:


KeenanRileyFaulkner (commented):
Currently, the phys_netlist tool can't correctly run without synthesis being done with `-flatten_hierarchy full -max_dsp 0` set. Those options look like they are no longer being passed to the synth tool in the appropriate flows, which could cause this PR to continue to fail CI even after the current "unexpected kwarg 'synth' passed to phys_netlist tool" problem is resolved. Everything else looks good to me.

PR #386, opened 2023-12-02
combine utils

yonnorc42:
- Moved ninja_utils into utils
- Combined bfasst/ninja_utils and bfasst/utils


yonnorc42:
This was failing the unittests locally so I'm surprised it passed

PR #401, opened 2023-12-07
create legacy tools

### LINKED ISSUE #315: Rename `ninja_tools` and `ninja_flows` to just `tools` and `flows`.  Create `_legacy` for older tools/flows

### LINKED ISSUE #402: Update make pylint target
Currently, the `make pylint` target checks all `.py` files in HEAD that differ from the common ancestor of HEAD and origin/main. This seems to be a relic of ensuring that the codebase need not be refactored all at once when pylint was first added to it. 

The CI checks handle pylint differently; they simply check all `.py` files at HEAD in the `scripts/` and `bfasst/` directories using the following command:
```
pylint --errors-only $(git ls-files --directory scripts --directory bfasst | grep -e ".py$")
```
And more broadly uses pylint on all files in the `env.GIT_DIFF_FILTERED` secret:
```
pylint ${{ env.GIT_DIFF_FILTERED }}
```

The `make pylint` target _[also does this](https://github.com/byuccl/bfasst/blob/ef507db2b6e351f349e78e45ce2c40f37172ecd3/Makefile#L124C2-L125C87)_ in addition to the comparison with origin/main. The downside of doing so is that it will error out if files are deleted from the working branch that still exist in main. Because the project has largely been made to follow pylint requirements the check between branches may be unnecessary at this point.

yonnorc42:
- Moved legacy tools to new directory tools_legacy
- Added new subdirectory of scripts called scripts_legacy 
- Changed pylint to no longer compare to origin/main
- Added new flow VivadoBitToNetlist
- Fixes #315 
- Fixes #402


yonnorc42:
@KeenanRileyFaulkner When I run pylint on this branch, I get this error:

`(.venv) yonnorc42@CCL2:~/bfasst$ make pylint
find ./scripts ./bfasst -iname "*.py" -exec black -q -l 100 {} \;
git fetch
pylint --errors-only $(git ls-files --directory scripts --directory bfasst | grep -e ".py$")
pylint $(git diff --name-only $(git merge-base origin/main HEAD) | grep -e ".py$")
************* Module bfasst/impl/__init__.py
bfasst/impl/__init__.py:1:0: F0001: No module named bfasst/impl/__init__.py (fatal)
************* Module bfasst/opt/__init__.py
bfasst/opt/__init__.py:1:0: F0001: No module named bfasst/opt/__init__.py (fatal)
************* Module bfasst/reverse_bit/__init__.py
bfasst/reverse_bit/__init__.py:1:0: F0001: No module named bfasst/reverse_bit/__init__.py (fatal)
************* Module bfasst/synth/__init__.py
bfasst/synth/__init__.py:1:0: F0001: No module named bfasst/synth/__init__.py (fatal)

------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 0.00/10, +0.00)`

The paths for those files has changed but pylint is still checking their old paths, which is causing the error, but I don't know why.

KeenanRileyFaulkner:
When you run pylint locally it compares any files you changed. Try running it again locally and see if the second time it gets the files from the correct location? Your pylint checks passed on this PR

yonnorc42:
It gets them from the wrong location still no matter how many times it's run.

KeenanRileyFaulkner:
Give me a minute. I'm looking at your branch and trying to fix it

KeenanRileyFaulkner:
This may be normal behavior, since you removed files. The pylint make target checks out the common ancestor of your current HEAD and the main branch and compares any files that are different between them with the pylint tool. In your case, you branched off the fix_weeklytests branch, but that's irrelevant. Even if you branched off main, pylint would still try to check files that are missing from your branch. However, the PR is passing those same checks on CI. I'm not 100% clear on why CI doesn't act exactly the same, but I think that once this merges and you pull again, it should go away locally.

KeenanRileyFaulkner:
Looking at the pylint workflow in the `.github/` directory, it looks like CI checks all the python files in your bfasst and scripts directories, but doesn't compare them to main, hence why it passes CI but not the `make pylint` target. This could be an issue worth opening, so that the make pylint target does the same thing.

jgoeders (changes requested):

.github/actions/setup_env/Makefile
```diff
@@ -8,7 +8,7 @@ install_fasm2bels:
 	cd ${BFASST_PATH_FASM2BELS} && make build
 	cd ${BFASST_PATH_FASM2BELS} && make test-py
 
-	cd ../../.. && python scripts/run_design.py designs/basic/and3/ xilinx_and_reversed
+	cd ../../.. && python scripts/scripts_legacy/run_design.py designs/basic/and3/ xilinx_and_reversed
```
jgoeders, line 11: Let's use run.py for this.

.github/workflows/unittests.yml
```diff
@@ -114,4 +114,4 @@ jobs:
   #     run: |
   #       source .venv/bin/activate
   #       source env.sh
-  #       python scripts/run_experiment.py ${{ matrix.experiment }} -j8 --print_period 0
+  #       python scripts/scripts_legacy/run_experiment.py ${{ matrix.experiment }} -j8 --print_period 0
```
jgoeders, line 117: Let's just remove this entire commented out section from this file.

Makefile
```diff
@@ -92,7 +92,7 @@ install_fasm2bels:
 	$(IN_ENV) cd third_party/fasm2bels && make test-py
 
 # Run a simple design through fasm2bels to generate the database.
-	$(IN_ENV) python scripts/run_design.py designs/basic/and3/ xilinx_and_reversed
+	$(IN_ENV) python scripts/scripts_legacy/run_design.py designs/basic/and3/ xilinx_and_reversed
```
jgoeders, line 95: Let's use run.py for this.

jgoeders (changes requested):

Makefile
yonnorc42: @jgoeders How do you use run.py to run that same flow? Or does it not matter which flow/design is used here?

KeenanRileyFaulkner: `xilinx_and_reversed` turned into `VivadoBitAnalysis`

KeenanRileyFaulkner: @yonnorc42 just for personal reference, the flow used here does matter in that it has to use f2b. The first time f2b runs, it generates a database for the part and its bitstream, which is used on subsequent runs of the tool.

yonnorc42:
@jgoeders I think this should be ready for merge

.github/actions/setup_env/Makefile
```diff
@@ -8,7 +8,7 @@ install_fasm2bels:
 	cd ${BFASST_PATH_FASM2BELS} && make build
 	cd ${BFASST_PATH_FASM2BELS} && make test-py
 
-	cd ../../.. && python scripts/run_design.py designs/basic/and3/ xilinx_and_reversed
+	cd ../../.. && python scripts/run.py VivadoBitAnalysis designs/basic/and3
```
jgoeders, line 11: Hmm, the VivadoBitAnalysis actually uses the PhysicalToLogical tool, which is private.  The CI has access to it, but the public does not.  Can you make a new flow `VivadoBitToNetlist` that stops after Xray, and use that here?

Jeff

Makefile
```diff
@@ -92,7 +92,7 @@ install_fasm2bels:
 	$(IN_ENV) cd third_party/fasm2bels && make test-py
 
 # Run a simple design through fasm2bels to generate the database.
-	$(IN_ENV) python scripts/run_design.py designs/basic/and3/ xilinx_and_reversed
+	$(IN_ENV) python scripts/run.py VivadoBitAnalysis designs/basic/and3
```
jgoeders, line 95: Same problem here.

jgoeders (changes requested):

.github/actions/setup_env/Makefile
yonnorc42: Do we need to add tests to this new flow? @jgoeders

PR #409, opened 2023-12-12
fix weekly_tests

yonnorc42:
- Removed design_dirs as it's reduntant
- Changed synth to synth_options because it was causing weeklytests to fail


jgoeders (approved):

tests/weekly_tests/phys_netlist_paper.yaml
```diff
@@ -64,4 +64,4 @@ designs:
 
 flow: vivado_phys_netlist_cmp
 
-synth: "-flatten_hierarchy full -max_dsp 0"
\ No newline at end of file
+synth_options: "-flatten_hierarchy full -max_dsp 0"
```
jgoeders, line 67: I think this is added by the flow already, and not required here.  Can you double check?

tests/weekly_tests/vivado_ooc.yaml
```diff
@@ -3,4 +3,4 @@ design_dirs:
 
 flow: vivado_ooc
 
-synth: "-flatten_hierarchy full"
\ No newline at end of file
+synth_options: "-flatten_hierarchy full"
```
jgoeders, line 6: I think this is added by the flow already, and not required here.  Can you double check?

jgoeders (changes requested):

tests/weekly_tests/phys_netlist_paper.yaml
yonnorc42: This one is added by the flow already

tests/weekly_tests/vivado_ooc.yaml
yonnorc42: I'm not actually sure on this one, vivado_ooc.py doesn't specify it's synth_options like vivado_phys_netlist_cmp.py does and I don't know how else to check.

tests/weekly_tests/vivado.yaml
```diff
@@ -1,4 +1,4 @@
-design_dirs:
+designs:
```
jgoeders, line 1: Are you sure this works?  I thought you needed `design_dirs` when you want it to iterate through all subdirectories.

tests/weekly_tests/vivado_ooc.yaml
```diff
@@ -1,6 +1,69 @@
...
+  - ooc/sudoku/
+  - ooc/tiny_encryption_algorithm/
+  - ooc/uart2spi/
+  - ooc/vga/
+  - ooc/wb_lcd/
+  - ooc/WB_stage/
 
 flow: vivado_ooc
-
-synth: "-flatten_hierarchy full"
```
jgoeders, line 6: If this isn't done in the flow, then this probably needs to come back.

jgoeders (changes requested):

tests/weekly_tests/vivado.yaml
yonnorc42: I thought it worked because the test actually runs unlike when it's `design_dirs` but I didn't notice it wasn't going into every design in each subdirectory. The reason I made this change in the first place because the test won't run with it as `design_dirs` because of an issue with how `design_dirs` works.

yonnorc42: Actually I think I misspoke before, I think using `designs` works fine. I made a change that made the test run using `design_dirs` and it did the same thing as when it was `design`

jgoeders: OK, didn't realize it worked.  I've opened #418

yonnorc42:
@jgoeders Do you know why the gmt_tools clone is failing? It says the host key verification is failing but I don't know why that would start happening now. It happened on my other pull request as well.

jgoeders:
@yonnorc42 I had to change how we are brining in private submodules in order to support brining them in one by one.  Some changes needed to be made on the runner computers, but I've just been fixing the runners as the error comes up.  If you see another, just message me and I'll fix that machine.  ...they might all be fixed now.

PR #411, opened 2023-12-15
update unittests

### LINKED ISSUE #408: vivado invocation logs
In our vivado [rules](https://github.com/byuccl/bfasst/blob/main/bfasst/tools/common/vivado_rules.ninja.mustache), the command is:

```
command = export tempjou=$$(mktemp); export templog=$$(mktemp); cd $cwd && {{ utils_path }}/retry.sh {{ vivado_path }} -mode batch -journal $$tempjou -log $$templog -source $in >&- && mv $$tempjou $journal && mv $$templog $log
```

I'm wondering why we are putting the logs in a temp directory. It's a bit annoying because you can't look at them while they are running.

Also the logs are only copied over if the tool runs correctly...

### LINKED ISSUE #421: Unittest has long run time.  Move long running tests to existing tests.
Unittest takes a long time to run.  This is mainly due to the tests that run the flows, then invoke ninja again to make sure there's no work to do.  This is a good test, but we could move this extra check to our other tests that run the flows.  At the end of running the CI yaml file, it could have another step that runs ninja to make sure there's no work to do.

yonnorc42:
- Added test for VivadoBitToNetlist flow
- unittest no longer runs any of the flows end-to-end
- run-test now checks if ninja has more work to do after the flow runs (except on vivado_phys_netlist_error_inject.yaml)
- vivado_ooc weeklytest now excludes 2 designs in ooc/ because "Some designs in the OOC directory have too many I/Os to be placed"
- vivado logs aren't put into temp directory and then moved over if tool succeeds anymore
- Adjusted quick_test_flows.sh (does anyone use this?)
Fixes #421 
Fixes #408


test/scripts/test_run.py
jgoeders, line 1: @yonnorc42 Can we just delete this entire file?  Is it still being used for anything?

yonnorc42: It checks that run.py works as expected when different arguments are passed in. It checks that run.py will work when given either a .yaml file or a flow and design, and it checks if run.py fails if any other arguments are passed in.

yonnorc42:
@jgoeders Do you know why conformal is failing to run?

jgoeders:
@yonnorc42 The license server was down after the flood.  It's back up now, but there seems to be some issue downloading rapidwright. I'm guessing the current error is temporary.

yonnorc42:
@jgoeders Is the issue with RapidWright just that we took out the recipe for it in the makefile?

PR #424, opened 2023-12-21
Took synth_options out of tests/weekly_tests/vivado_ooc.yaml

yonnorc42:
The VivadoOoc flow doesn't expect any synth options. It errors if you try to give any. I'm not sure if that means it doesn't need any synth options, or if that needs to be added to the flow. @jgoeders @KeenanRileyFaulkner I'm not 100% clear on the effects "flatten_hierarchy full" has so some clarification would be nice.

### DIFF
--- a/tests/weekly_tests/vivado_ooc.yaml
+++ b/tests/weekly_tests/vivado_ooc.yaml
@@ -1,6 +1,5 @@
 designs:
   - ooc/
 
-flow: vivado_ooc
+flow: VivadoOoc
 
-synth: "-flatten_hierarchy full"


KeenanRileyFaulkner:
Yeah, the ooc flow doesn't need synth options. [Here's an explanation of the flatten hierarchy switch.](https://support.xilinx.com/s/article/52257?language=en_US)

yonnorc42:
Thanks. I don't think it will need it because the CI check for vivado_ooc doesn't use it and it passes.

KeenanRileyFaulkner:
👍 that might be because they are running different designs though. I'm not sure.

PR #429, opened 2024-01-15
part database

### LINKED ISSUE #415: Xray Dependency on part database
I ran into an interesting issue where I deleted my fasm2bels and then ran `make install_fasm2bels` to reinstall it.  At the end of the install steps it runs a flow to regenerate the part database; however, since I had run the flow previously, I got `ninja: no work to do.`

The `XRay` tool should probably be updated to add the part database as a dependency so that ninja throws an error if the file does not exist.

yonnorc42:
- Guarantees that flow that generates part database (for fasm2bels) runs even if the flow already ran previously
Fixes #415 
@jgoeders What do you think about just doing this?

### DIFF
--- a/scripts/install-fasm2bels.sh
+++ b/scripts/install-fasm2bels.sh
@@ -49,11 +49,13 @@ else
 	cd ${BFASST_PATH_FASM2BELS}
     make env
 	make build
-	make test-py
     cd -
 
     # Run a design to generate the part database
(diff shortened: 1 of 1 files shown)


jgoeders:
I rather not copy the build directory to another location as I don't want to end up with many files lying around for the user.  I'm wondering if we can somehow check for this case another way.

yonnorc42:
@jgoeders When exactly do you want to check for the parts database then? Do we want to check that it exists after install_fasm2bels runs or when we're running a flow that uses xray?

PR #439, opened 2024-02-01
ip dump

dallinjdahl:
- adding ip dumping information
- fixing typo


KeenanRileyFaulkner (changes requested):
You will also need to remove "labelfile" related lines in [the tool](https://github.com/byuccl/bfasst/blob/bf06d9542308b21a6154025a608117f9eeb82f74/bfasst/tools/transform/randsoc_dump.py#L21C9-L21C76) and [the flow](https://github.com/byuccl/bfasst/blob/6aa210eea3d8dedc797aa6cd694984be6feb2b6e/bfasst/flows/rand_soc_dumped.py)

KeenanRileyFaulkner (approved):

PR #447, opened 2024-02-21
Fix randsoc assertion when no control signal to connect

jgoeders:


dallinjdahl:
Are we concerned with these failing checks?

jgoeders:
No, they have something to do with the fasmb2bels virtual environment and look unrelated.  I have no idea why they just popped up today, as I'm not sure what's changed to cause this issue.

They shouldn't affect the randsoc generation.

PR #452, opened 2024-03-06
Const nets fix

yonnorc42:
If multiple const nets are found, the first one is used, and the rest have their connections moved over to the first net and then they're removed. Fixes a specific design failing.

@reillymck As of now I haven't added in the code that removes the extra GND nets and cells, but this works. Will those extra nets and cells even show up anywhere if they're not connected to anything?


reillymck:
I'm not sure. If it does show up it will be in the verilog file generated in phys netlist. Will you do a search for gnd primitives and verify there is only one?

yonnorc42:
@reillymck There's only one.

reillymck (approved):

PR #455, opened 2024-03-17
netlist_cmp fixes

### LINKED ISSUE #334: Some Phys Netlist Cmp designs do not work with Vivado 2023.2
See designs here: https://github.com/byuccl/bfasst/blob/main/experiments/xilinx_phys_netlist_cmp.yaml

yonnorc42:
- Remove netlist cleanup from netlist cmp, passes without it
- Ignore CASCADEIN and CASCADEOUT pins in net equivalence step
- Ignore certain pins _before_ checking if they're not yet mapped, this helps prevents issues during equivalence checks
- Lots of minor changes in structural comparison to prevent problems and unneeded operations
- Reverted to old version of get_pin method which returned a GND_PIN by default
Fixes #334 
@reillymck thoughts?


reillymck:
We can't ignore CASCADEIN pins. We need to make sure they aren't part of the design. Fasm2bels doesn't support cascading brams but they will still give some output. So when equivalence fails because CASCADE pins exist we need to know that's the reason and it's not because the tool isn't working right.

reillymck:
And we do need to make sure that both designs have zero/ground on the cascade pins. Also I'm not sure we can just ignore net equivalency because it's possible the ground signal connecting to them is generated by a LUT

yonnorc42:
@reillymck 

> We can't ignore CASCADEIN pins. We need to make sure they aren't part of the design. Fasm2bels doesn't support cascading brams but they will still give some output. So when equivalence fails because CASCADE pins exist we need to know that's the reason and it's not because the tool isn't working right.

What do you mean by "We need to make sure they aren't part of the design"? 

"So when equivalence fails because CASCADE pins exist we need to know that's the reason and it's not because the tool isn't working right." So do we want the designs with the CASCADE pins to fail? Should I not worry about these designs?

reillymck:
Sorry. They can be part of the design but we need to make sure the value on the pin is zero or ground.

yonnorc42:
I have no idea why pylint is mad but this should be ready to merge otherwise

yonnorc42:
@jgoeders Thanks. This is ready for merge now.

PR #461, opened 2024-04-02
parts database

### LINKED ISSUE #415: Xray Dependency on part database

yonnorc42:
Fixes #415

If there's a better way to do this, let me know. I thought this was an easy way to ensure the database is generated during the fasm2bels install.


yonnorc42:
I forgot the runners don't quite work the same way as the lab computers.

PR #462, opened 2024-04-04
synplify ninja

KeenanRileyFaulkner:
- got synplify to run
- updated readme and reverted conformal login info


KeenanRileyFaulkner:
This flow seems to cause conformal to fail on designs like `byu/alu` but it passes on `basic/and3`.

KeenanRileyFaulkner:
The CI is failing with external tool install issues, but otherwise this PR should be good to go.

jgoeders (approved):

PR #467, opened 2024-04-09
ninja_mode

### LINKED ISSUE #451: Pass -j to ninja
The run.py should have a way to pass the '-j' argument to ninja.  Sometimes you need to do a serial run...

yonnorc42:
- Added the option to run ninja in serial mode
Fixes #451


jgoeders:
Any reason to not directly expose the -j argument?  I think it might be nice to give the user control over the number of threads.

jgoeders:
@yonnorc42 The `-j` notation is actually fairly universal (it's what regular makefiles use as well), so in this case, I think it's best to follow convention. Thanks for making the change.

PR #485, opened 2024-06-07
Dump only bel information

dallinjdahl:
Making output of dump script more like bitstream.  Previously, the unique identifier was based on the cell, and we output the cell name as a key property.  Now by default the dump script outputs only things relating to bels.

### DIFF
--- a/bfasst/utils/dump.tcl
+++ b/bfasst/utils/dump.tcl
@@ -29,19 +29,6 @@ proc property {b p prefix out} {
 	
 proc common_properties {b out} {
 
-	# LUT6_2 returns two bels for the single cell C
-	if {[llength $b] > 1} {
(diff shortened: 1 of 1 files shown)


dallinjdahl:
This works, but neurosub isn't equipped to handle it yet.  It is a breaking change to the input format.  Even updating neurosub, we'd have to blow away our current datasets and regenerate them.

The advantage is that it's now all bels, all the way down.  I'm not sure it would change much in our results, but this is the most accurate bel-dump I know of right now.

dallinjdahl:
supporting pr in neurosub is https://github.com/byuccl/neurosub/pull/21

jgoeders:
@dallinjdahl I'm going to merge this and update the datasets

PR #486, opened 2024-06-10
Pure bels

dallinjdahl:
We now work exclusively with bels, and GND and VCC are gone.  Hopefully the model should still cope well.


jgoeders:
@dallinjdahl It looks like this is failing.  Do you know why?

dallinjdahl:
This should be good to go now.

Issue #488, opened 2024-06-11
randsoc_dump dependency


jgoeders:
There seems to be a missing dependency causing randsoc_dump to fail sometimes.

This will create `design_0/design.tcl`:
```
build /home/jgoeders/bfasst/build/rand_soc/design_0/design.tcl: rand_soc | /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/paths.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/utils.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ports.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/creator.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/__init__.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/emc.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/slice_and_concat.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/accumulator.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/ip_base.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/gpio.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/axi.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/uartlite.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/microblaze.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/intc.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/rand_soc/ip/clk_gen.py /home/jgoeders/bfasst/third_party/gmt_tools/rand_soc/main.py
    design_dir_path = /home/jgoeders/bfasst/build/rand_soc/design_0
    seed = 0
    part = xc7a200tlffv1156-2L
```

However, the output file is not a dependency for the subsequent synthesis step:
```
[...]

PR #489, opened 2024-06-11
Fix for #488

reillymck:
Makes sure that the tcl files called by run.tcl are dependencies to run.tcl being created.


reillymck:
Also, what are the chances that pr #489 is the fix for issue #488

PR #490, opened 2024-06-11
remove rule to build itself

reillymck:
Our first ninja rule in every file was to rebuild the master ninja build file. I think this caused the following:
image
I'm pretty sure that ninja does not expect or handle the build file changing while it is executing it?
Ninjas job output "[jobs_complete/total_jobs]" for me seemed to be one less than the number of build statements in the file (for doing a single design). 
Anyway, I think once the build file was rewritten, ninja starts the build count over again. Also, sometimes it replaces the current line, so you do not notice that it is happening, and sometimes it creates a new line like shown in the image.

### DIFF
--- a/master.ninja.mustache
+++ b/master.ninja.mustache
@@ -1,11 +1,3 @@
-rule configure
-    command = python {{ FLOWS_PATH }}/ninja_flow_manager.py $flow $in --flow_arguments "$flow_arguments"
-    generator = 1
-
-build build.ninja: configure {{#designs}}{{ . }} {{/designs}} | {{#deps}}{{ . }} {{/deps}}
-    flow = {{ flow }}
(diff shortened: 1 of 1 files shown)


reillymck:
Everything done by these lines in the build.ninja file is done before ninja is ran [here](https://github.com/byuccl/bfasst/blob/2e4d3f990b22457d56b503d91bd42f1c68a6de12/scripts/run.py#L56C9-L59C1)

reillymck:
Fails unittests because there is now one less build statement for every ninja file. But I notice that the unittest specifically checks for the configure rule. 
I see that the rule is marked as a generator, which means ninja re-invokes itself. So I think this is actually a feature, but I'm not sure why.

PR #491, opened 2024-06-18
Added a caching mechanism for the possible_matches dict

yonnorc42:
For the larger designs like jpegencode, a lot of the runtime is spent initializing the list of possible matches for each primitive based on cell type and properties. This pr creates a caching mechanism which rebuilds the dict based on the instance names. For jpegencode, this brought the runtime down by 29 minutes. This only works if you don't change the design at all in between runs.


yonnorc42:
This is broken right now

yonnorc42:
This works now

yonnorc42:
Uh that was on accident

PR #492, opened 2024-06-21
added caching mechanism

yonnorc42:
For the larger designs like jpegencode, a lot of the runtime is spent initializing the list of possible matches for each primitive based on cell type and properties. This pr creates a caching mechanism which rebuilds the dict based on the instance names. For jpegencode, this brought the runtime down by 29 minutes.

### DIFF
--- a/bfasst/utils/structural.py
+++ b/bfasst/utils/structural.py
@@ -27,13 +27,13 @@ class StructuralCompare:
     """Structural compare and map"""
 
     def __init__(
-        self, named_netlist_path, reversed_netlist_path, log_path, use_cache=False
+        self, named_netlist_path, reversed_netlist_path, log_path, debug=False
     ) -> None:
         self.reversed_netlist_path = reversed_netlist_path
         self.named_netlist_path = named_netlist_path
         self.named_netlist = None
         self.reversed_netlist = None
-        self.use_cache = use_cache
+        self.debug = debug
 
         self.log_path = log_path
         logging.basicConfig(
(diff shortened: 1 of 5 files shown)


yonnorc42:
Error injection has failed twice in a row. It's been failing every once in a while so it's odd that it happened twice in a row. If it happens a third time, it's maybe some from my pull request, but I'm not sure what would cause that.

bfasst/utils/structural.py
```diff
@@ -277,47 +323,87 @@ def map_ports(self):
...
+                    raise StructuralCompareError(
+                        f"Not equivalent. {named_instance.name} \
+                        has no possible match in the netlist."
+                    )
 
-            self.possible_matches[named_instance] = instances_matching.copy()
+                self.possible_matches[named_instance] = instances_matching.copy()
+
+            possible_matches = {}
+            # pylint: disable=consider-using-dict-items
```
reillymck, line 388: Why did you decide to disable this check? Maybe I'm missing why dict items wasn't quite working.

bfasst/utils/structural.py
```diff
@@ -22,11 +30,12 @@ class StructuralCompareError(Exception):
 class StructuralCompare:
     """Structural compare and map"""
 
-    def __init__(self, named_netlist_path, reversed_netlist_path, log_path) -> None:
+    def __init__(self, named_netlist_path, reversed_netlist_path, log_path, debug=False) -> None:
```
reillymck, line 33: Did you change other files to use this debug parameter? Or when you use the debug feature how are you setting it?

yonnorc42: I've been using a script that runs comparison directly without using ninja to allow me more flexibility and customization. Should I add it in as a flow argument?

reillymck: I think what we should do is have it dump the debug file by default, but only look for if the file exists if it is passed in as a command line argument. Maybe --debug should  tell it to look for the file, and it dumps the file regardless. How does that sound?

yonnorc42: That sounds good to me.

yonnorc42:
I decided to do it as a flow arg, by default it's false but if you add --flow_arguments "{'debug': True}" to the end of your "python scripts/run.py ... " it will use the cache if there is one available, and use the RapidWright functionality as well. If debugging mode is False, and/or if there is no cache available, it will create a cache during that run.

reillymck (approved):

PR #495, opened 2024-07-12
handle dsps

yonnorc42:
Handles DSPs by ignoring the pin if it meets the criteria of being the opposite const net of the original, and the IS_x_INVERTED property being true for that pin. Let me know what you think @reillymck If this solution isn't viable, then we need to adjust some fasm2bels code instead of our code, because this is the only solution using our code that I've thought of.

### DIFF
--- a/bfasst/utils/structural.py
+++ b/bfasst/utils/structural.py
@@ -357,12 +357,10 @@ class StructuralCompare:
         logging.info("Considering %s (%s)", instance.name, instance.cell_type)
 
         # Get the implemented potential instance to map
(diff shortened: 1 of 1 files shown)


yonnorc42:
I need to fix this, it's not working right now

yonnorc42:
This is working now

reillymck:
I'm not sure if we can do this since the property isn't showing up for one of the designs. For all we know fasm2bels is right about the property and vivado is hiding it. Let's hold off on putting this in main for now. But, let's keep using it for now so that these pins pass and we see if there are other things to work on or if this is the only thing holding us back from passing.

PR #497, opened 2024-07-26
Speed up Comparison

yonnorc42:
I'm starting this pull request now but it's going to be a draft for a while until I've figured out exactly what changes we want to make and how to make them.

### DIFF
--- a/bfasst/utils/structural.py
+++ b/bfasst/utils/structural.py
@@ -509,7 +509,13 @@ class StructuralCompare:
             logging.info("===== Mapping Iteration %s =====", iteration)
 
             # Loop through reversed netlist blocks
-            instance_iter = iter(set(self.named_netlist.instances_to_map))
+            # Sort the instances based on the length of their possible matches list
+            sorted_instances = sorted(
+                set(self.named_netlist.instances_to_map),
+                key=lambda instance: len(self.possible_matches[instance])
+            )
+            # Create an iterator from the sorted instances
+            instance_iter = iter(sorted_instances)
             try:
                 while not overall_progress:
                     overall_progress = self.potential_mapping_wrapper(next(instance_iter))


yonnorc42:
One of the changes I've made might dramatically slow down a run the first time, but should increase the speed in subsequent runs since the list of possible matches for each instance should hopefully be significantly shorter... as long as I get it working.

yonnorc42:
While using the cache, the runtime of jpegencode through comparison is about 16 minutes.

bfasst/utils/structural.py
```diff
@@ -693,11 +707,14 @@ def check_for_potential_bram_mapping(self, named_instance):
                     raise StructuralCompareError("Unexpected BRAM CASCADE Configuration")
                 logging.info("Unexpected BRAM CASCADE Configuration for %s", named_instance.name)
 
-        instances_matching_connections = [
-            i for i in self.possible_matches[named_instance] if i not in self.block_mapping.inverse
-        ]
-
-        for pin in named_instance.pins:
+        instances_matching_connections = self.possible_matches[instance_name] - set(
```
reillymck, line 710: Is there a significant cost to having to cast self.block_mapping.inverse to a set?

reillymck:
I was thinking about what you said in the meeting about the dictionary lookup times to for `{instance_name: instance}` dictionary. Since this datatype does not change, we should use a list. The potential matching sets should just have the list indices for the master instances list (instead of using the instance name keys).. Python list access should be [O(1)](https://docs.python.org/3/glossary.html#term-list)
What do you think?

yonnorc42:
Looking at the profile, I don't think it is actually taking as long as I thought. It was only a quarter of a second in the entire jpegencode run.
@reillymck 
`3244748    0.246    0.000    0.246    0.000 {method '__getitem__' of 'dict' objects}`

bfasst/utils/structural.py
yonnorc42: I don't think so, but I haven't been able to find a cast method or __init__ method for set creation in the profile, so I can't actually see. I think we need to break this file into a lot more functions. It would make for a more accurate profile because then we would really be able to tell exactly which calls are taking up the most time.

yonnorc42:
Also, from what I'm seeing online, dict access is also about O(1)

bfasst/utils/structural.py
yonnorc42: For example, in check_for_potential_mapping and check_for_potential_bram_mapping, we have this redundant code:

`tmp = {
                instance
                for instance in instances_matching_connections
                if self.reversed_instance_map[instance].get_pin(name, idx).net == other_net
            }
        if not tmp:
            if other_net.is_gnd:
                tmp = {
                    instance
                    for instance in instances_matching_connections
                    if self.reversed_instance_map[instance].get_pin(name, idx).net.is_gnd
                }
            elif other_net.is_vdd:
                tmp = {
                    instance
                    for instance in instances_matching_connections
                    if self.reversed_instance_map[instance].get_pin(name, idx).net.is_vdd
                }`

reillymck (changes requested):
It is looking pretty good. Can you upload the latest profile results?

dictionary [] access should be __get_item__. I looked and I think the get() calls are spydrnet. 
Dictionary access is O(1), since the time it takes to hash the key doesn't change based on the size of the datatype. But index access of course would be less instructions. However, I do not think we are impacted by this enough to care based on the __get_item__ profile.

bfasst/utils/structural.py
```diff
@@ -321,99 +333,90 @@ def map_ports(self):
...
-        # assert not instance.cell_type.startswith("SDN_VERILOG_ASSIGNMENT")
-
-        logging.info("Considering %s (%s)", instance.name, instance.cell_type)
+    def potential_mapping_wrapper(self, instance_name: str) -> bool:
+        """Wrap check_for_potential_mapping some inital checks/postprocessing"""
+        instance = self.named_instance_map[instance_name]
+        logging.info("Considering %s (%s)", instance_name, instance.cell_type)
 
         # Get the implemented potential instance to map
         if not instance.cell_type.startswith("RAMB"):
```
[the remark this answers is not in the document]
yonnorc42: I think to change this, we'd have to add a lot of redundant code. All of the functions in the program use instances_to_map. For example, here https://github.com/byuccl/bfasst/blob/0762a069e06b9d5ca627cbda033b04362b7c6431/bfasst/utils/structural.py#L599
We'd have to do a string comparison or something similar to know which list it needs to be removed from, so I think the way it is structured right now might actually be best.

yonnorc42:
Going to push a few minor netlist_cleanup changes as well, same logic slightly different format

yonnorc42:
Just added the flow argument logging_level to these tools, netlist_cleanup, error injector, phys netlist, and structural comparison. 
Example flow run:
python scripts/run.py VivadoPhysNetlistCmp --flow_arguments "{'logging_level': 'ERROR'}" 
Will only log messages with logging.error or above, so no logging.debug, logging.info, or logging.warning messages will get added to the log file

yonnorc42:
This should be good to merge @reillymck or we can talk about it during the meeting if there's anything else to be added

reillymck (approved):

PR #501, opened 2024-08-14
Check logging for f-strings

reillymck:
The logging library functions allow for kwargs to process "%" strings. Our coding style is f-strings, and this will have pylint check the logging calls to enforce this.


reillymck:
Update I think we want to keep the old formatting. This way strings will only be formatted if the log is printed.

PR #514, opened 2024-10-31
Dataset metrics

KeenanRileyFaulkner:
Adds two scripts to bfasst, allowing for efficient computation of metrics on randsoc-based graphs for gnn training datasets. The metrics can be computed ad-hoc through the use of command line options. Bfasst is used to iterate through a dataset, run process_graph.py on each graph in it, then run accumulate_metrics.py at the end.

If things run for too long, and you don't want to wait for all instances of process_graph.py to finish, kill the run and invoke the accumulation script on the output directory directly.

The final output is contained in a file called `summary_stats.log`, with individual metrics for each instance of each ip contained in `master_metrics.log`. accumulate_metrics.py will overwrite these files every time, but options are available to specify a different output name, if the utility is used directly.


KeenanRileyFaulkner:
Needs dependencies on the scripts in the ninja build statements, still, so that rebuild triggers when utils update

bfasst/flows/analyze_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+            graph_metrics_tool.create_build_snippets()
+
+        AccumulateMetrics(self, pieces).create_build_snippets()
+
+    @classmethod
+    def flow_build_dir_name(cls) -> str:
+        """Get the name of the build directory for this flow"""
+        return "dataset_metrics"
+
+    def add_ninja_deps(self, deps):
```
jgoeders, line 44: I don't think this is required if it's just calling super.  The behavior will be the same if you remove it.

bfasst/utils/accumulate_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+
+                # Concatenate the lists
+                master_metrics[ip][metric].extend(values)
+
+    return master_metrics
+
+
+def sort_metrics(metrics):
+    """Sort the values for each metric in the dictionary."""
+    for ip, _ in metrics.items():
```
jgoeders, line 101: If you want to loop through keys, just use:
```
for ip in metrics:
```

jgoeders (approved):

jgoeders:
@KeenanRileyFaulkner It looks like this is failing CI.  Anything I can help with?

KeenanRileyFaulkner:
Sorry. I didn't realize there was a formatting issue. I'm out of town I'm right now, but will try to fix it when I have a chance this weekend.

KeenanRileyFaulkner:
@jgoeders All tests are passing now

PR #516, opened 2024-12-05
Mem fix

reillymck:
Currently only tested on the byu/ designs.


yonnorc42:
I adjusted the make_matches_by_nets function, did a little bit of refactoring for pylint, and added in a helper file for structural.py to move some of the functions into to make the file a lot smaller and easier to read (haven't moved much over yet)

yonnorc42:
Replaced names with indices in the for all of the data structures. Haven't tested the memory difference yet but it's working

PR #517, opened 2025-01-08
koios adjustments

yonnorc42:
- Update koios test to add designs that work with kintex board and leave additional comments
- Adjust koios design to use less IO
- Add part override to design yaml files (review this to check if the way I implemented it is fine, in both yaml_parser.py and flow.py?)


yonnorc42:
Should be working now

reillymck (approved):

PR #519, opened 2025-02-12
cleanup

yonnorc42:
- Bumped spydrnet to a new branch which doesn't rename constant nets, we already have logic to deal with that in bfasst and the spydrnet method for renaming the constants is extremely slow.

- Use remove_children_from instead of remove_child function in the cleaner. The remove_child function has extremely slow repeated list removal operations. remove_children_from creates a copy of the list with the children that need to be removed excluded, which ends up being much faster.  

For reference, in the main branch, it took jpegencode 204 minutes (3 hrs 24 minutes) to clean. On this branch, it took 26.5 minutes.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -77,7 +77,7 @@ jobs:
 
     - name: 'Upload artifacts'
       if: failure()
-      uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


reillymck:
This branch is merged into mem_fix

PR #525, opened 2025-02-21
output interchange files

reillymck:


reillymck:
Physical netlist generator outputs a logical capnp netlist in its build directory "phys_logical_netlist.capnp"
Fasm2bels outputs capnp objects to its build directory *.capnp
Logical netlist schema for the datatype is found [here](https://github.com/chipsalliance/fpga-interchange-schema/blob/c985b4648e66414b250261c1ba4cbe45a2971b1c/interchange/LogicalNetlist.capnp)
Rapidwright example for loading a capnp logical netlist is found [here](https://github.com/Xilinx/RapidWright/blob/master/src/com/xilinx/rapidwright/interchange/LogicalNetlistExample.java)

PR #532, opened 2025-05-07
Trying to fix weekly

jgoeders:


yonnorc42:
I moved my changes from the fix_clock branch to here since it also has a patch for an issue causing a weekly test to fail. The weekly tests haven't finished yet, but I think everything except the rand_soc test will pass.

yonnorc42:
The weekly tests installed fasm2bels successfully so it looks like this is working now. I'm keeping the fasm2bels cache in the weekly tests since it takes a long time, but if there's a new fasm2bels version, the install should work fine. I added the cmake installation to the make install rule as well.

All of the weekly tests are passing now except for rand_soc, which I've never really looked at before so I don't know what the problem is.

reillymck (approved):

PR #534, opened 2025-05-08
Fix clock

yonnorc42:
If you add in a clock to the design.yaml file of a design, Vivado performs unwanted physical optimizations that cause some designs to fail comparison. This change just tells Vivado not to use the clock unless you run the ClockCrank flow.


yonnorc42:
Closing this pr and adding changes to the fix-weekly pr

PR #537, opened 2025-05-19
VivadoReimplCompare Flow to Compare Two Implemented Designs

EddieRydell:
Created a new flow, VivadoReimplCompare, which:
- Synthesizes a design
- Runs Vivado implementation
- Changes the synthesis netlist in some way (netlist_transform.py)
- Runs Vivado implementation again on the new netlist
- Compares the two implementations for physical differences (physcmp.py)

Also created two new tools and associated utils:
- NetlistTransform, which obfuscates LUT INIT values and saves the original values to a file for later restoration
- PhysCmp, which compares the physical layout, timing, utilization, and power reports of implemented designs


jgoeders:
@reillymck Can you review this and give feedback to @EddieRydell 

@EddieRydell It looks there are some files that shouldn't be committed as part of this.  Can you remove them?  Also, please work toward fixing it so that the tests all pass.

bfasst/flows/flow_descriptions.yaml
```diff
@@ -81,6 +81,12 @@ flows:
   - fasm2bels
   - rapidwright
 
+- name: VivadoReimplCompare
```
jgoeders, line 84: Maybe let's call this flow something else?  We should come up with a cool name for the project.

PR #538, opened 2025-05-20
ImplObfuscate

EddieRydell:
New flow to synthesize a design, obfuscate the netlist, implement it, and compare the obfuscated netlist with the original netlist


jgoeders:
@EddieRydell Can you add a CI test for your flow?

bfasst/utils/netlist_obfuscate.py
```diff
@@ -0,0 +1,208 @@
...
+
+def obfuscate_lut(inst):
+    """
+    Overwrite INIT in LUT if present.
+    Returns True if changed.
+    """
+    ref_name = inst.reference.name
+    try:
+        lut_size = int(ref_name.replace("LUT", ""))
+    except ValueError:
```
reillymck, line 30: What is causing a value error here? Is it when there is a LUT6_2?

bfasst/utils/physcmp.py
```diff
@@ -0,0 +1,251 @@
...
+
+def load_and_trim(path: Path) -> list[str]:
+    with open(path, "r") as f:
+        return f.readlines()[11:]
+
+
+def compare_reports(golden_path: Path, test_path: Path, label: Optional[str] = None):
+    """Run a text-based comparison between two implementation reports."""
+
+    def is_rise_fall_swap(gs: str, ts: str) -> bool:
```
reillymck, line 54: I think we should avoid declaring functions inside other functions.

bfasst/utils/physcmp.py
```diff
@@ -0,0 +1,251 @@
...
+        saw = True
+        logging.debug("[%s] Test has %d extra lines.", label, len(extra_t))
+
+    return saw
+
+
+def compare_cells(d1: Design, d2: Design) -> int:
+    """Compare physical placement of all cells in a design"""
+
+    by_name2 = {c.getName(): c for c in d2.getCells()}
```
reillymck, line 114: See Design.getCell(String name)

reillymck (changes requested):

bfasst/utils/netlist_obfuscate.py
EddieRydell: Yeah I think it was. I refactored to deal with LUT6_2 separately now.

bfasst/utils/netlist_deobfuscate.py
```diff
@@ -0,0 +1,188 @@
...
+    """Setup logging for this util"""
+
+    level = getattr(logging, level_str.upper(), logging.INFO)
+    logging.basicConfig(
+        filename=log_path,
+        level=level,
+        format="%(asctime)s %(levelname)s: %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+    )
+    console = logging.StreamHandler()
```
reillymck, line 35: This is good for debugging, but for main we want output going to files.

PR #542, opened 2025-06-17
Refactor ImplObfuscate

EddieRydell:
Modified flow so now more cell properties are obfuscated. Both designs are also now run through RapidWright. We verify full bitstream equivalence at the end.
I also changed the flow so that opt_design is run before obfuscation. This avoids having to deal with some optimizations that are hard to restore.


reillymck:
I see you have set all the DRC Checks to warnings. Do you know what DRCs are failing?
Do you know if this works without flattening the design before implementation?

reillymck:
Rapidwright output should be redirected to a log file as shown [here](https://github.com/byuccl/bfasst/blob/e0f0d15842a2709736d560eb023d010bf7ea143e/bfasst/utils/rw_phys_netlist.py#L82)

EddieRydell:
I had to do some messy stuff to get rid of the interchange files so unfortunately everything is now in one commit. All the weekly tests should also pass now.

PR #553, opened 2025-07-07
Update black

reillymck:
I found a bug in the black version we were using where it failed to produce an output that did not change the functionality of my code. It has been patched in the latest black version.

Update formatting of repo files to use the new black. Update requirements.txt


reillymck:
This is branched off of retime. Merge PR #552 first and then sync this PR with main to see the actual changes from adding black.

PR #555, opened 2025-07-14
Added flow_run command line function and autocomplete for flow names

EddieRydell:
This PR adds a new flow_run function that serves as a shorthand for `python scripts/run.py`. It includes Bash autocompletion for flow names, and still allows standard file or argument completion after a flow is selected.

The autocompletion is implemented via a helper script and a new setup_autocomplete.sh installer. The setup script now is also invoked when calling `make install` or `make setup_autocomplete`. 

Note: These changes probably won't work in a shell that isn't bash.


reillymck:
I checked out the branch, ran make setup_autocomplete, and then sourced the environment. However, it is not autocompleting flow names for `$python scripts/run.py` I also tried making run.py executable and running `$./scripts/run.py` with no also no result.

reillymck (approved):

reillymck (approved):

